PR #29, opened 2015-01-09
Add zfs_import_dir kernel parameter

jasonrm:
What are your thoughts on allowing the zpool import directory to be set via a kernel parameter?

I need this (or something like it) because I've been unable (export with correct paths, correct zpool cache file) to get my boot pool to import by partuuid via the hook. Especially important when I change hardware configurations (such as when I added cache and log vdevs).

I've verified a correct zpool cache file is in the initcpio image, and done the import/export dance from archiso. Nothing has resulted in it importing by partuuid except for this modification.

kernel parameters examples:

`zfs_import_dir=/dev/disk/by-id/`
`zfs_import_dir=/dev/disk/by-uuid/`
`zfs_import_dir=/dev/disk/by-partuuid/`

### DIFF
--- a/zfs-utils-git/zfs-utils.initcpio.hook
+++ b/zfs-utils-git/zfs-utils.initcpio.hook
@@ -1,4 +1,5 @@
 ZPOOL_FORCE=""
+ZPOOL_IMPORT_FLAGS=""
 
 zfs_get_bootfs () {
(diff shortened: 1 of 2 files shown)


techmunk:
Was there a reason these flags were not added to the run_latehook() ?

For example, if /var or /usr, with mountpoint's of legacy are used as ZFS roots, then you would not be able to use this feature for them?

Was this intentional?

PR #32, opened 2015-02-01
Avoid absolute paths in hooks

rwanyoike:
This makes us more resilient against path changes. The shell is smart enough to resolve the path for us.


rwanyoike:
Abandoned

PR #33, opened 2015-02-01
Support bootfs child datasets

rwanyoike:
Currently only the dataset defined by the `zfs` kernel parameter will get mounted during boot. For example (`zfs=bootfs`):

``` 
$ zfs list -o name,mountpoint
NAME               MOUNTPOINT
zroot              /            # (bootfs)
zroot/home         /home
zroot/usr          legacy
zroot/var          legacy
```

The other datasets will wait until the `zfs.target` systemd service is called (IF it is enabled). A dataset like `zroot/usr` will cause the kernel to complain that it can't find `/sbin/init` (`/sbin -> /usr/bin`).

Similar issue: https://github.com/demizer/archzfs/issues/22

---

These changes try to mount any *children*, with the dataset defined by the `zfs` kernel parameter. For example (`zfs=bootfs`):

```
$ zfs list -o name,mountpoint
NAME               MOUNTPOINT   
zroot/root         /            # (bootfs)
zroot/root/usr     legacy       # will get mounted (child)
zroot/root/var     legacy       # will get mounted (child)
zroot/scratch      none         # will not get mounted
zroot/home         /home        # will not get mounted
```

Legacy mountpoints get mounted if they are listed in /etc/fstab.

### DIFF
--- a/zfs-utils-git/zfs-utils.initcpio.hook
+++ b/zfs-utils-git/zfs-utils.initcpio.hook
@@ -20,7 +20,6 @@ zfs_get_bootfs () {
 }
 
 zfs_mount_handler () {
-    local node=$1
     if [ "$ZFS_DATASET" = "bootfs" ] ; then
         if ! zfs_get_bootfs ; then
             # Lets import everything and try again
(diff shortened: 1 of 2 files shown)


rwanyoike:
Abandoned

PR #40, opened 2015-05-14
Prevent 'ash: 1: unknown operand' error when zfs_force option is not set

techmunk:


demizer:
Thanks for making this. Unfortunately it got fix in the initscript refactor before I had a chance to merge this.

Issue #39, opened 2015-05-08
Add a fully ZFS-enabled GRUB package


kerberizer:
The current Arch [GRUB packages](https://www.archlinux.org/packages/?name=grub) don't recognize ZFS pools upgraded to support the most recent feature flags. This is because [upstream GRUB](http://git.savannah.gnu.org/cgit/grub.git) lacked proper support for the `hole_birth` and `embedded_data features` up until very recently.[1,2]

The problem doesn't affect booting, i.e. even systems with unpatched GRUB and fully upgraded pools boot normally, but grub-install would fail to recognize the ZFS filesystem on the required devices. This essentially makes new installations with ZFS next to impossible.

Seeing that it took nine months for the required patches to get incorporated upstream, and not to mention that we'll have to wait even more for a new GRUB release and then updated Arch packages, it makes sense for us to maintain a separate GRUB package with always current support for ZFS.  

**NB:** Pull request #41 follows progress.

---
1. http://git.savannah.gnu.org/cgit/grub.git/commit/?id=a1007c6af296172c788f3346ff3ad30b9020f723
2. http://git.savannah.gnu.org/cgit/grub.git/commit/?id=4a7ea4003bd9f39e9bf2ebadbe4ab38f8874be66

techmunk:
It'd be awesome to have a grub package built and available with the other ZFS packages.

I've been able to get grub working 100% with arch by taking the following steps: (This is for a whole disk ZFS on root install)

#### Can be done on any machine to build the ZFS enabled grub package
1. Download Grub git package from the AUR. https://aur.archlinux.org/packages/grub-git/
2. Install zfs-git package on build machine. (Adds ZFS include files for building Grub)
3. Add `--enable-libzfs` to configure lines in grub-git PKGBUILD
4. makepkg

#### Install steps from ZFS enabled archiso image.
1. See https://wiki.archlinux.org/index.php/Installing_Arch_Linux_on_ZFS for main info, but essentially follow on from "Setup the ZFS filesystem". **For whole disk setup DO NOT DO THE "Partition the destination drive" STEP**.
2. After your `pacstrap /mnt base zfs-git`, but before chrooting into /mnt, be sure to copy the grub package compiled above, into this new root.
[...]

PR #41, opened 2015-05-15
DO NOT MERGE: Add the grub-zfs package

kerberizer:
### DO NOT MERGE!

This is WiP on issue #39. The package will also need to be integrated into the build infrastructure and it might be better to update it to a newer tagged GRUB code revision where the ZFS-related patches have already been merged upstream.


techmunk:
Nice. It would be amazing to get a grub-zfs package up and running.

Do you not need the --enable-libzfs configure flag? I had issues when /boot on ZFS would not get recognised unless I used that configure flag. I might do some messing around on the weekend to see if that really is the case. I also ran the script at https://github.com/zfsonlinux/grub over the zfs and spl include files prior to building. Though I'm really not sure if I needed that. Really my main goal is to not require another partition for /boot, which I have with the steps I outlined in https://github.com/archzfs/archzfs/issues/39#issuecomment-102209883

kerberizer:
This is now badly outdated and doesn't support some of the latest ZFS pool feature flags. The ordinary [grub](https://www.archlinux.org/packages/core/x86_64/grub/) official package and [grub-git](https://aur.archlinux.org/packages/grub-git/) from AUR should both offer much better ZFS support.

Issue #53, opened 2016-04-07
Support bootfs child datasets (Patch)


rwanyoike:
Currently only the dataset defined by the `zfs` kernel parameter will get mounted during boot. For example (`zfs=bootfs`):

``` 
$ zfs list -o name,mountpoint
NAME               MOUNTPOINT
zroot              /            # (bootfs)
zroot/home         /home
zroot/usr          legacy
zroot/var          legacy
```

The other datasets will wait until the `zfs.target` systemd service is called (IF it is enabled). A dataset like `zroot/usr` will cause the kernel to complain that it can't find `/sbin/init` (`/sbin -> /usr/bin`).

Similar issue: #22

---

These changes try to mount any *children*, with the dataset defined by the `zfs` kernel parameter. For example (`zfs=bootfs`):

```
$ zfs list -o name,mountpoint
NAME               MOUNTPOINT   
zroot/root         /            # (bootfs)
zroot/root/usr     legacy       # will get mounted (child)
zroot/root/var     legacy       # will get mounted (child)
zroot/scratch      none         # will not get mounted
zroot/home         /home        # will not get mounted
```

Legacy mountpoints get mounted if they are listed in /etc/fstab. :+1:

rwanyoike:
As per the submitting changes (https://github.com/archzfs/archzfs/tree/495964c#submitting-changes):

[child_datasets_1-3.txt](https://github.com/archzfs/archzfs/files/209316/child_datasets_1-3.txt) - [PATCH] Mount the root, and any non legacy child datasets
[child_datasets_2-3.txt](https://github.com/archzfs/archzfs/files/209317/child_datasets_2-3.txt) - [PATCH] Mount the root, and any child datasets
[child_datasets_3-3.txt](https://github.com/archzfs/archzfs/files/209318/child_datasets_3-3.txt) - [PATCH] Avoid absolute paths in hook

demizer:
Awesome. Thank you!

ajs124:
Why has this not been merged yet? My system does not boot without this patchset.

demizer:
@ajs124 because time

But tonight's the night.

Issue #49, opened 2015-11-15
Longer wait for /dev/zfs inside zfs-utils.initcpio.hook


Bronek:
On machines with many or large ZFS pools it can take more than 10s to start ZFS . Currently `zfs-utils.initcpio.hook` is limited to waiting 10s which may result in failure to boot the system . This limit should be increased to 30s (or perhaps configurable value)

```
diff --git a/zfs-utils.initcpio.hook b/zfs-utils.initcpio.hook
index c1a4d00..00948d4 100644
--- a/zfs-utils.initcpio.hook
+++ b/zfs-utils.initcpio.hook
@@ -85,8 +85,8 @@ run_hook() {
         echo "ZFS: No hostid found on kernel command line or /etc/hostid. ZFS pools may not import correctly."
     fi

-    # Allow up to 10 seconds for zfs device to show up
-    for i in 1 2 3 4 5 6 7 8 9 10; do
+    # Allow up to 30 seconds for zfs device to show up
+    for i in `seq 30`; do
         [ -c "/dev/zfs" ] && break
         sleep 1
     done
```

Obviously md5sum of `zfs-utils.initcpio.hook` should be updated as appropriate inside PKGBUILD

Bronek:
Nice, thank you!

PR #57, opened 2016-05-09
Rewrite the initcpio hook

dasJ:
This PR ensures a lot of variables are corretly quoted, closes some PRs, but also drops some legacy stuff.

#### Things that got better
- Variables are now quoted and use the `${}` syntax
- The script got shorter
- Support for `ro` when using `bootfs`
- Correct use of `warning` and `die`, these are functions from `mkinitcpio`
- Use of bash tests (`[[` instead of `[`), as this is recommended by the `mkinitcpio` project ([reference](https://git.archlinux.org/mkinitcpio.git/tree/HACKING))
- Mount subdatasets (see #53) of the root
- Allow customizing the timeout for `/dev/zfs` (see #49)
- Fixes the error message from #49

#### Dropped backwards compatibility
- `zfs` kernel parameter dropped
- ZFS root is now mounted with `root=zfs=bootfs` or `root=zfs=dataset/whatever`
- Messages like `Importing pool zroot` are removed, I'll add them later

### DIFF
--- a/src/zfs-utils/zfs-utils.initcpio.hook
+++ b/src/zfs-utils/zfs-utils.initcpio.hook
@@ -60,36 +60,26 @@ zfs_mount_handler () {
 run_hook() {
     # Force import the pools, useful if the pool has not properly been exported
     # using 'zpool export <pool>'
-    [[ $zfs_force == 1 ]] && ZPOOL_FORCE='-f'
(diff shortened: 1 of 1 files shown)


dasJ:
This needs some more work

demizer:
Still some useful stuff in here. Thanks!

dasJ:
Note that this is bash code and does not work on the busybox sh.

demizer:
:+1: @dasJ

PR #65, opened 2016-08-16
add libgcc_s.so.1 to dependencies (fixes #64)

### LINKED ISSUE #64: Created initrd is missing libgcc_s.so.1
Upon upgrading to 0.6.5_r403_ge35c5a8_4.7.0_1-1 of zfs-linux-git, my system fails to boot, with a message:
```
libgcc_s.so.1 must be installed for pthread_cancel to work
```
and the zpool command then aborts. I get similar behavior if I try to manually import the pool from the initrd shell.

I was able to work around this by copying libgcc_s.so.1 into /usr/lib within the initrd, and now the system boots.

grmblfrz:


demizer:
Thanks!

Issue #100, opened 2017-01-28
Force rootfs mountpoint


OrdinaryMagician:
I use a particular dual-boot setup (which works in Gentoo but not in Arch) where I have one root dataset for each distro, with mountpoints that are NOT /.
```
NAME             USED  AVAIL  REFER  MOUNTPOINT
Yukari           793G   106G   658G  /media/Data
Yukari/Alice    2.70G   106G  2.70G  -
Yukari/Arch     17.4G   106G  14.9G  /media/Arch
Yukari/Games     106G   106G   105G  /media/Data/Games
Yukari/Gentoo   5.63G   106G  5.63G  /media/Gentoo
Yukari/Mugetsu   124M   106G   124M  -
Yukari/Temp      248K   106G   152K  /tmp
```

At boot, Gentoo pretty much ignores the mountpoint of the root dataset and it just becomes / automatically, while Arch just gets mounted at /media/Arch. But, due to how the Arch initcpio hook works, it tries to mount at /new_root/media/Arch, fails, and I get dropped to a shell. I hope it's not asking much to get an extra option for this case so it gets mounted where it should.

johnramsden:
Are you trying to mount your dataset to root '/'? If you are you need to set set ```zfs set mountpoint=/ <pool>/<dataset>```. 

Maybe I'm misunderstanding you. 

If you're trying to mount it somewhere else and it's unable to mount, it's possible you have something in the directory. It won't like mounting the dataset unless there is an empty mount point.

OrdinaryMagician:
I want it to mount at / without setting the mountpoint to /. That's all I'm saying. It's as simple as not appending the mountpoint to "/new_root" if the dataset being mounted is the root dataset.

Kiwwiaq:
Than I would set all desired mount points as legacy and let `/etv/fstab` do the job.

OrdinaryMagician:
The thing is, I cannot use a legacy mountpoint in this case, because it's the root. You need to mount the root before there's an /etc/fstab available. And if no /etc/fstab is available, all datasets with legacy mountpoints are ignored, and so, booting fails yet again.

OrdinaryMagician:
I figured out a simple way to fix my specific case without affecting others' setups, I'll open a pull request in a bit, after I test it works fine.

OrdinaryMagician:
Closing this since my pull request was merged.

PR #106, opened 2017-02-17
Strip mountpoint from rootfs if it is not "/"

OrdinaryMagician:
This easily allows for multi-distro setups where each system has its own root dataset with separate mounts (e.g.: /media/Arch, /media/Gentoo, etc.).

Additionally it strips the root mountpoint from any children, so separate /usr and others will still work as expected.


OrdinaryMagician:
This is a fix for my issue #100

demizer:
Thank you for the PR! I will test this once I have some spare cycles soon. It looks fine, but I don't want to break anyone's root fs, so I have to be careful here.

PR #123, opened 2017-06-15
Update zfs_initcpio_install_hash

hexchain:
Following #119. Sorry for not updating the hashes in `conf.sh` because I've only just realized it.


demizer:
Thanks again!

PR #126, opened 2017-07-08
add support for native zfs encryption in initcpio hook

minextu:
This will enable installations on encrypted datasets. I only tested passphrases as key format.

Becomes functional, as soon as https://github.com/zfsonlinux/zfs/pull/5769 is merged upstream. 
In the meantime https://aur.archlinux.org/packages/zfs-encryption-dkms-git/ can be used for testing.


demizer:
Adding WIP to subject as we are waiting for upstream to merge feature. Once that happens I will merge this into mainline.

zdykstra:
https://github.com/zfsonlinux/zfs/commit/b52563034230b35f0562b6f40ad1a00f02bd9a05

demizer:
@zdykstra thanks

It appears zfs encryption could be available in the git packages at least.

src/zfs-utils/zfs-utils.initcpio.hook
```diff
@@ -25,6 +25,27 @@ zfs_get_bootfs () {
     return 1
 }
 
+zfs_decrypt_fs() {
+    dataset=$1
+    
+    # check if 'zfs load-key' is available
+    zfs 2>&1 | grep load-key > /dev/null || return 0
```
minextu: This will check if encryption is available, so this change should also work on stable

minextu:
This can safely be merged, it won't do anything on stable

demizer:
Thanks everyone!

PR #128, opened 2017-07-11
Add linux-hardened

minextu:
Depends on https://github.com/archzfs/archzfs/pull/127, closes https://github.com/archzfs/archzfs/issues/117

I only enabled the stable ZFS version for now. AUR packages already exist (https://aur.archlinux.org/packages/zfs-linux-hardened/)

PR #127, opened 2017-07-11
update to OpenZFS version 0.6.5.11

minextu:
This version does compile with gcc 7.1. So packages can finally be updated!

This fixes https://github.com/archzfs/archzfs/issues/121, fixes https://github.com/archzfs/archzfs/issues/122 and fixes https://github.com/archzfs/archzfs/issues/118

PR #129, opened 2017-07-12
Simultaneous installation of multiple zfs modules

minextu:
Depends on: https://github.com/archzfs/archzfs/pull/127, https://github.com/archzfs/archzfs/pull/128

I separated `zfs-utils` and `spl-utils` from their kernel version, and renamed them to `zfs-utils-common` and `spl-utils-common`. 
This allows to install multiple zfs kernel modules for different kernels, but only if they are using the same zfs version. So you could install `zfs-linux`, `zfs-linux-lts` and `zfs-linux-hardened` together, but you can't mix git and non-git zfs versions.

In order to archive this, I also had to create header packages for zfs and spl.
All old utils packages have been deleted as submodule, since their are not needed anymore.
The following new packages will have to be submitted to the aur: `zfs-utils-common`, `zfs-utils-common-git`, `spl-utils-common` and `spl-utils-common-git`
All header packages will be created automatically, once the new PKGBUILDs are submitted.

I know, that you already worked on something like that, so don't hesitate on rejecting this merge request.
In contrast to your implementation in #extramodules (which I couldn't get to work), headers are generated using the PKGBUILD of zfs and spl, so ~~no~~ only minimal modification to build.sh were made, but I did create a new kernel config named `common`.


minextu:
Update: I added support for building extramodules in https://github.com/minextu/archzfs/tree/extramodules
This is based on this pull request and some of your work (https://github.com/archzfs/archzfs/commit/8f4d83b1559d8e076f374ec79aff9fc79adaf7ea)

As said previously: If you would rather continue your version of extramodules, don't hesitate on closing this merge request. I do this mostly because I'm learning Bash and more about Arch Linux along the way

minextu:
I just created a test repo, with all theses changes. Just replace archzfs with this:
```
[zfs-test]
SigLevel = Never
Server = https://arch.et.tc/$repo
```

demizer:
❤️ awesome work!

demizer:
Once I merge this, we will need to update the Arch Wiki docs

Issue #92, opened 2016-12-18
Use extramodules to eliminate package rebuilds for minor kernel releases


demizer:
Chris sent an email and made a comment on the AUR zfs-linux page informing me that he was able to get the packages working using extramodules; same as the nvidia kernel modules. This will eliminate the need to rebuild the packages for minor kernel releases (4.8.8 -> 4.8.9). I had attempted this four years ago and was not able to get it to work due to upstream support and forgot about it. This would be a major improvement for everyone.

storrgie:
Can you reference this email in some way so that other people might poke into it? It seems like this might be one of the crux issues that generates a lot of tickets for the project.

demizer:
@storrgie absolutely. Email was sent to me by @lnxsrt last October.

> I posted a comment in the zfs-linux AUR about using a dependency of linux>=4.8 instead of linux=4.8.2 so that you don't have to make packages for each kernel point release. Many packages (nvidia, etc) use this method along with the /lib/modules/extramodules-4.8 folder that will remain consistent after a point release kernel update. I have attached a .tgz with the 4 primary zfs-linux packages that have been modified to ignore point releases. It works fine for me, I will note that I made the version 0.6.5.8_4.8.999_1 just to keep yaourt from trying to grab your package from the AUR. Ideally you could start this on 4.9 and just call it 0.6.5.8_4.9_1. Let me know if it would prefer me to fork and submit a merge request on your github.

He also provided an attachment with the changes. I will make a branch and reference it here. Apologies for having this in the backlog for so long, I have been extremely busy at work, but I plan on tackling this this weekend because kernel updates are coming so quickly now and it's painful for all of us.

Thanks!

minextu:
Doesn't work correctly (see #130)

PR #130, opened 2017-07-14
[WIP] Build extramodules

minextu:
Depends on: https://github.com/archzfs/archzfs/pull/129
Closes https://github.com/archzfs/archzfs/issues/92

I added support for building extramodules, which will allow to target multiple point releases of a kernel without having to update zfs all the time. So if the linux package for example gets update from 4.11.2 to 4.11.3, you don't have to wait for an updated package in archzfs.

This is based on some of your work (8f4d83b).
As said previously: If you would rather continue your version of extramodules, don't hesitate on closing this merge request. I do this mostly because I'm learning Bash and more about Arch Linux along the way

At this point, this is experimental, since while updating from linux-lts-4.9.36 to 4.9.37 did work flawlessly, an update from linux-hardened 4.12.e to 4.12.1.a did break unfortunately. Might be a problem with linux-hardened though.

I did also create a test repo, with all theses changes
Just replace the archzfs one with this:
```
[zfs-test]
SigLevel = Never
Server = https://arch.et.tc/$repo
```
You might need to "downgrade" some packages with pacman -Syuu, since the version numbers are lower.

### DIFF
--- a/src/kernels/linux-hardened.sh
+++ b/src/kernels/linux-hardened.sh
@@ -4,8 +4,9 @@ mode_desc="Select and use the packages for the linux-hardened kernel"
 
 # Kernel versions for hardened packages
 pkgrel="1"
-kernel_version="4.13"
-kernel_version_max=$(echo ${kernel_version}| awk -F. '{print $1"."$2+1}')
+kernel_version="4.13.3"
+kernel_version_full=$(kernel_version_full ${kernel_version})
+kernel_version_max=$(echo ${kernel_version_full}| awk -F. '{print $1"."$2"."$3+1}')
 
 # Kernel version for GIT packages
 pkgrel_git="${pkgrel}"
@@ -37,9 +38,9 @@ header="\
 
 update_linux_hardened_pkgbuilds() {
     pkg_list=("spl-linux-hardened" "zfs-linux-hardened")
-    kernel_mod_path="extramodules-${kernel_version}-hardened"
+    kernel_version_major=${kernel_version_full%\.*}
+    kernel_mod_path="extramodules-${kernel_version_major}-hardened"
     archzfs_package_group="archzfs-linux-hardened"
-    kernel_version_full=$(kernel_version_full ${kernel_version})
     spl_pkgver=${zol_version}.${kernel_version_full}
     zfs_pkgver=${zol_version}.${kernel_version_full}
     spl_pkgrel=${pkgrel}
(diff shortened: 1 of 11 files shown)


demizer:
WOW! You are amazing. Thanks so much! I have been super busy at work this week so haven't had much time to merge PRs and test. I should be able to get to it later tonight or this weekend.

src/kernels/common.sh
```diff
@@ -0,0 +1,81 @@
+# For build.sh
```
demizer: I like this. 👍

src/kernels/linux-hardened.sh
```diff
@@ -0,0 +1,109 @@
...
+    spl_src_target="https://github.com/zfsonlinux/zfs/releases/download/zfs-${zol_version}/spl-${zol_version}.tar.gz"
+    zfs_src_target="https://github.com/zfsonlinux/zfs/releases/download/zfs-${zol_version}/zfs-${zol_version}.tar.gz"
+    spl_workdir="\${srcdir}/spl-${zol_version}"
+    zfs_workdir="\${srcdir}/zfs-${zol_version}"
+    linux_depends="\"linux-hardened>=${kernel_version}\" \"linux-hardened<${kernel_version_max}\""
+    linux_headers_depends="\"linux-hardened-headers>=${kernel_version}\" \"linux-hardened-headers<${kernel_version_max}\""
+    zfs_makedepends="\"${spl_pkgname}-headers\""
+}
+
+# update_linux_hardened_git_pkgbuilds() {
```
demizer: do we not want git builds for the linux-hardened kernel?

src/kernels/linux-lts.sh
```diff
@@ -4,11 +4,13 @@ mode_desc="Select and use the packages for the linux-lts kernel"
 
 # Kernel versions for LTS packages
 pkgrel="1"
-kernel_version="4.9.36-1"
+kernel_version="4.9"
```
demizer: This makes me so happy. 😄

demizer:
I see it has WIP in the subject. Are we good to merge this @minextu ?

minextu:
We might need more testing before merging this, since the package for linux-hardened broke for the last update. I don't know how this will affect future kernel releases

src/kernels/linux-hardened.sh
minextu: I didn't want to submit zfs-linux-hardened and zfs-linux-hardened-git to the aur before knowing, if this would get merged 😄 
You can just uncomment it, and it should work

minextu:
Update from linux-hardened 4.12.1.b to 4.12.2.a did break again, while 4.12.1.a to 4.12.1.b did work. So I will assume, that this is a problem with linux-hardened and use a lower range of versions for this kernel.

minextu:
Packages build for linux-lts `4.9.37` don't work with `4.9.40`. I will continue testing, but it seems like this is not stable enough and we will have to rebuild zfs for every minor version unfortunately. 

We could however target a linux kernel without the pkver, for updates from `linux-4.12.1-1` to `linux-4.12.1-2`. That would be better than nothing...

minextu:
I changed this patch, to only support a pkgrel increase. So something like `4.12.10-1` is compatibly with `4.12.10-2`. A bigger version range is too unstable...

Issue #137, opened 2017-07-27
ZFS on Linux 0.7.0 released


fryfrog:
Looks like 0.7.0 released a handful of hours ago.

https://github.com/zfsonlinux/zfs/releases/tag/zfs-0.7.0

fryfrog:
And @demizer is clearly aware, working with ZFS on Linux to resolve at least one issue...

https://github.com/zfsonlinux/zfs/issues/6409

demizer:
Thanks for reporting the issue.

Just to keep everyone up-to-date, there are those missing manual pages, and I am also having issues running my tests because I believe some zpool command line arguments have changed.

I will try to get this out as soon as I can, but I am only able to work on this two hours a day as I am just getting back from vacation and work starts back up tomorrow. I am hoping the issues with my tests are simple to fix.

Thanks!

minextu:
I you want to help testing: I update my zfs-test repo to the latest version:
```
[zfs-test]
SigLevel = Never
Server = https://arch.et.tc/$repo
```
This does also contain the extramodules patch (https://github.com/archzfs/archzfs/pull/130).

I will try to get the test scripts running, so I can help with testing in the future

PR #138, opened 2017-07-27
set OpenZFS version 0.7.0

minextu:
Closes https://github.com/archzfs/archzfs/issues/137

I got the test scripts working, and had no errors. I ran `archzfs-qemu-lts-test-00-bootfs` and `archzfs-qemu-std-test-00-bootfs`
I could start the resulting images and login with no problems.


demizer:
Thank you!

PR #154, opened 2017-07-31
more improvements

### LINKED ISSUE #146: Split common into common-git
common-git packages need to be updated separately from stable common packages. When building the common package group using build.sh, the stable common packages are also built. When going to add the newly built packages to the repo using repo.sh, only the common-git packages are need but the stable packages are added as well. This might cause conflicts if the signatures are not the same for users that have already downloaded the packages previously.

Requirements:
* Add common-git package group to build.sh, repo.sh, and push.sh
* When releasing git packages, only the common-git packages are added to a repo using repo.sh and push.sh

### LINKED ISSUE #148: build.sh: do not sign packages
Package signing is only needed when adding packages to a repository. Signing should be moved to repo.sh

Requirements:

* build.sh does not sign anything

* repo.sh signs packages before adding them to the target repository

minextu:
fixes #146, fixes #148


demizer:
Great work! Thank you.

PR #155, opened 2017-08-01
Update hardened packages for kernel 4.12.4.a-1

minextu:
Would be great, if you add linux-hardened to the archzfs repo


demizer:
I will get these added to the repo and the scraper scripts. Thanks!

PR #181, opened 2017-09-05
Fix build.sh

minextu:
An `exit 1` has been added in 9136ed6c6f811d2c6dfe977989d5ff8c41f28a30. This prevents building more than one package per run.


demizer:
Oops!

PR #193, opened 2017-12-03
Workaround for linux 4.14

### LINKED ISSUE #191: Update SPL (Git?) to 0.7.4-pre to include support for 4.14 kernels
The latest commit, as of this writing:

zfsonlinux/spl@ed19bccfb651843fa208232b3a2d3d22a4152bc8

... is required to support 4.14 kernels, as they de-exported vfs_read/vfs_write, and added new functionality to existing functions, so now SPL either uses the new functionality, or falls back to wrappers that use the older functions.

My AUR kernel has already jumped onto 4.14, so I'm running a release behind for the time being, but it probably won't be long before the Arch kernel package updates to some 4.14 release.

I assume this will have to make it into the Git package? The version on spl-dkms-git and zfs-dkms-git looks suspiciously behind, unless I am to understand that 0.7.3 was never tagged, or the version string doesn't use `git describe --tags` to versionize the packages, so it displays as 0.7.0+somehugeoffset, instead of 0.7.3+slightlysmalleroffset.

E: Oh yeah, saw your other notice. See you tomorrow at the earliest, and enjoy your break!

### LINKED ISSUE #192: ZFS Module Fails to Load
After rebooting my machine, the zfs file system is not loaded. When I try to load the module manually, I am greeted with:

```
lithium% sudo modprobe zfs
modprobe: ERROR: could not insert 'zfs': Unknown symbol in module, or unknown parameter (see dmesg)
```

I am working to resolve the issue.

minextu:
fixes #192, fixes #191 

If you still don't have a solution: here is a workaround.
Just revert this commit, as soon as a stable zfs release contains this patch.

I haven't run testing/test.sh though...

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -157,6 +157,9 @@ generate_package_files() {
         run_cmd_no_output "source ${script_dir}/src/spl/PKGBUILD.sh"
         msg2 "Creating spl.install"
         run_cmd_no_output "source ${script_dir}/src/spl/spl.install.sh"
+        msg2 "Copying linux4.14.patch"
+        run_cmd_no_output "cp ${script_dir}/src/spl/linux4.14.patch ${spl_pkgbuild_path}/linux4.14.patch"
+
 
         msg2 "Creating zfs PKGBUILD"
         run_cmd_no_output "source ${script_dir}/src/zfs/PKGBUILD.sh"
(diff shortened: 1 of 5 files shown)


minextu:
I installed this locally though, and it works on my system. So this should be safe to merge

demizer:
Awesome. Thanks! I have done the same. I will merge this PR with my own changes.

kode54:
Now to wait for this to make it into the dkms packages, since I use an unsupported kernel. (linux-lqx)

PR #189, opened 2017-11-27
Fix dependency to fixed common-utils versions

minextu:
Fixes #188 

@demizer, you would still need to manually push all `git` packages at once, but this will disallow the installation of mismatched versions if they get into the repo.

However: It will make it harder for people, who build the packages manually using the aur, since they need to install `zfs-utils-common` and `zfs-<kernel>` in one operation to not break the dependency.
```
error: failed to prepare transaction (could not satisfy dependencies)
:: zfs-linux: installing zfs-common (0.7.2-1) breaks dependency 'zfs-common=0.7.1'
```

PR #195, opened 2017-12-04
Fix dependency for a fixed common-utils version (git packages)

minextu:
Part 2 of https://github.com/archzfs/archzfs/pull/189
Totally forgot about the git packages last time, sorry

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -561,7 +561,7 @@ check_zol_version() {
 check_mode() {
     # $1 the mode to check for
     debug "check_mode: checking '$1'"
-    
+
     # add all available modes
     if [[ "${1}" == "all" ]]; then
         for m in "${mode_list[@]}"; do
@@ -767,7 +767,7 @@ check_skip_build() {
             return 0
         fi
     fi
-    
+
     return 1
 }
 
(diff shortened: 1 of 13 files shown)


minextu:
Since we have a `pkgver()` function we get another problem by merging this: AUR users will only be able to build the package as long as no new commit as been added to zfs.
When pkgver() updates the version, the version of  the common-git dependency doesn't update automatically with it...

Maybe it would be better to not merge this. With the addition of an `all` command, out of sync packages might not happen anymore and aur users could still build the packages without modifying the pkgbuild

demizer:
Another thing we could do is lock the git checkout to a specific commit for people building from AUR. This commit target is updated with automation if we rebuild the git packages using the all or common-git modes of build.sh, which happens often. This way users can build git packages from aur without breakage. This also makes it so when we run tests against the git packages, we know for sure the checkout works.

demizer:
We do something similar at https://github.com/archzfs/archzfs/blob/master/build.sh#L303

minextu:
I added new commits to implement that

demizer:
Thank you!

Issue #194, opened 2017-12-03
Add "all" command to build scripts


demizer:
The "all" command would build all the packages (if there are changes).

For example, common-git often needs to be updated because of new changes in the zfs or spl repo. However, since we merged #189, all of the git packages need to be updated if common-git is updated. 

We could do this with:

```
./build.sh all update make -u
```

This command will:

1. build common and common-git (if there are changes).
1. build std, lts, hardened, dkms, and zen packages if there are change

and the same would be done with repo.sh and push.sh

PR #197, opened 2017-12-04
Allow multiple modes in build scripts

minextu:
Fixes https://github.com/archzfs/archzfs/issues/194

This does also add an `all` mode, to build all modes except for `archiso`.
I can't test push.sh. Should work fine though.


demizer:
Awesome! Thank you very much!

Issue #198, opened 2017-12-08
Update version informaton for *-git packages


UweSauter:
As of today (2017-12-08) all *-git packages in archzfs have wrong versioning information 0.7.0 while other packages reflect correct version as can be seen in the output below.

I'm not sure how to handle this case where version is a moving target (HEAD respectively the commit hash) but I believe that it should be updated to include the latest official release of ZFSonlinux, e.g. instead of `archzfs/spl-linux-git 0.7.0.r21.ged19bcc.4.14.3.1-4` it should be `archzfs/spl-linux-git 0.7.3.ged19bcc.4.14.3.1-4`.

```
# pacman -Ss zfs | grep ^arch
archzfs/spl-archiso-linux 0.7.2_4.13.3.1-1 (archzfs-archiso-linux)
archzfs/spl-linux 0.7.3.4.14.3.1-4 (archzfs-linux)
archzfs/spl-linux-git 0.7.0.r21.ged19bcc.4.14.3.1-4 (archzfs-linux-git)
archzfs/spl-linux-hardened 0.7.3_4.14.3.a.1-1 (archzfs-linux-hardened)
archzfs/spl-linux-hardened-git 0.7.0.r21.ged19bcc.4.14.3.a.1-1 (archzfs-linux-hardened-git)
archzfs/spl-linux-lts 0.7.3.4.9.66.1-2 (archzfs-linux-lts)
archzfs/spl-linux-lts-git 0.7.0.r21.ged19bcc.4.9.66.1-2 (archzfs-linux-lts-git)
archzfs/spl-linux-zen 0.7.3.4.14.3.1-3 (archzfs-linux-zen)
archzfs/spl-linux-zen-git 0.7.0.r21.ged19bcc.4.14.3.1-3 (archzfs-linux-zen-git)
archzfs/spl-utils-common 0.7.3-1 (archzfs-linux)
archzfs/spl-utils-common-git 0.7.0.r21.ged19bcc-2 (archzfs-linux-git)
archzfs/zfs-archiso-linux 0.7.2_4.13.3.1-1 (archzfs-archiso-linux)
archzfs/zfs-archiso-linux-headers 0.7.2_4.13.3.1-1
archzfs/zfs-linux 0.7.3.4.14.3.1-4 (archzfs-linux)
archzfs/zfs-linux-git 0.7.0.r200.gd4677269f.4.14.3.1-4 (archzfs-linux-git)
[...]

PR #200, opened 2017-12-08
Use date of the latest commit as version for git packages

minextu:
Closes https://github.com/archzfs/archzfs/issues/198

This will change versions of all git packages from something like 
`0.7.0.r200.gd4677269f.4.14.3.1-4`
 to
 `2017.11.15.r1059.gd4677269f.4.14.3.1`


demizer:
Thank you

PR #216, opened 2017-12-27
Support for Linux 4.15

### LINKED ISSUE #202: Patch copy should not run if there are no patches
This is not bad, but doesn't look good. If there are no patches then the cp command shows an error in the output:

```
++++ Copying spl-utils patches (if any)

Running command:

cp /home/demizer/src/arch/archzfs/src/spl-utils/*.patch packages/common-git/spl-utils-common-git/

cp: missing destination file operand after 'packages/common-git/spl-utils-common-git/'
Try 'cp --help' for more information.
Command returned: 1
```

We should not run the patch copy command if there are no patches.

### LINKED ISSUE #204: Remove untracked files from package directories
With ZFS On Linux version 0.7.4, the 0001-Linux-4.14-compat-vfs_read-vfs_write.patch patch is no longer needed, but it still exists in the packages directory which is uploaded to AUR.

Package directories should only contain files defined in the PKGBUILD.

minextu:
Early pull request for when kernel 4.15 hits arch.

Also fixes #202 and fixes #204

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -136,38 +136,43 @@ generate_package_files() {
 
     # Finally, generate the update packages ...
     if [[ "${kernel_name}" == "common" ]] || [[ "${kernel_name}" == "common-git" ]]; then
+        msg2 "Removing old spl-utils patches (if any)"
+        run_cmd_no_output "rm -f ${spl_utils_pkgbuild_path}/*.patch"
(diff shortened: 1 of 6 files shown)


minextu:
Replaced by https://github.com/archzfs/archzfs/pull/224

PR #217, opened 2017-12-27
Do not skip sha256sum when using multiple commands

minextu:
The spl/zfs src hash was always set to "SKIP", even on non git packages.


demizer:
Thanks!

PR #226, opened 2018-02-07
Add config files to backup array

### LINKED ISSUE #225: zed.rc replaced with default on upgrade of zfs-utils-common
It appears that anytime there is a update for the zfs-utils-common* packages the file `/etc/zfs/zed.d/zed.rc` is replaced with the default provided.

Could you please update the according zfs-utils-common* PKGBUILD's to include:
```
backup=('etc/zfs/zed.d/zed.rc')
```
As per the detail [here](https://wiki.archlinux.org/index.php/PKGBUILD#backup).

> **backup**
> An array of files that can contain user-made changes and should be preserved during upgrade or removal of a package, primarily intended for configuration files in /etc.
> 
> Files in this array should use relative paths without the leading slash (/) (e.g. etc/pacman.conf, instead of /etc/pacman.conf).
> 
> When updating, new versions may be saved as file.pacnew to avoid overwriting a file which already exists and was previously modified by the user. Similarly, when the package is removed, user-modified files will be preserved as file.pacsave unless the package was removed with the pacman -Rn command.

I would have submitted a PR for this, however, due to my lack of experience with submodules I can't seem to follow the existing submodules to see the said PKGBUILD's which require updating.

minextu:
Closes #225

### DIFF
--- a/src/zfs-utils/PKGBUILD.sh
+++ b/src/zfs-utils/PKGBUILD.sh
@@ -24,6 +24,7 @@ provides=("zfs-utils")
 install=zfs-utils.install
 conflicts=(${zfs_utils_conflicts})
 ${zfs_utils_replaces}
+backup=('etc/zfs/zed.d/zed.rc' 'etc/default/zfs')
 
 build() {
     cd "${zfs_workdir}"


demizer:
Thank you very much!

PR #238, opened 2018-04-04
Improve predictability at boot time and close #50

### LINKED ISSUE #50: Do not import all pools autmatically inside zfs-utils.initcpio.hook
Automatically importing all pools inside `zfs-utils.initcpio.hook` causes problems, for example loss of disk names inside `zpool.cache`, pool corruption on multipath SAN systems where the same pools might be already used by a different system at the same time, or slow boot times (and possibly pool corruption) if pools imported contain ZVOLs which themselves are ZFS pools (e.g. setup for virtual machines). This has been extensively discussed in ZOL at https://github.com/zfsonlinux/zfs/pull/3800 and https://github.com/zfsonlinux/zfs/issues/3777 and the agreement (with some reservations) is that ZOL should import pools from `zpool.cache` only, rather than import all pools. Archzfs already correctly installs systemd service `zfs-import-cache.service` which performs this without problems. Thus `zpool import -aN` should be removed from `zfs-utils.initcpio.hook` . This is what patch below does

```
diff --git a/zfs-utils.initcpio.hook b/zfs-utils.initcpio.hook
index 00948d4..80f8a2a 100644
--- a/zfs-utils.initcpio.hook
+++ b/zfs-utils.initcpio.hook
@@ -93,8 +93,4 @@ run_hook() {
 }


-run_latehook () {
-    /usr/bin/zpool import -N -a $ZPOOL_FORCE
-}
-
```

Obviously md5sum of `zfs-utils.initcpio.hook` should be updated inside PKGBUILD as appropriate

oowi:
- fix a problem introduced ~2018-03-18 where waiting for /dev/zfs
   to appear no longer waits for all, or enough, of the vdevs.

 - always use /etc/zfs/zpool.cache if zfs_import_dir is not specified,
   which generally makes imports faster.

 - obey command line arguments in the late hook.

 - mitigate issue #50:  when zfs_import_dir is not specified, do
   not, in the late hook, import any pool not mentioned in
   /etc/zfs/zpool.cache as of mkinitcpio time.

### DIFF
--- a/src/zfs-utils/zfs-utils.initcpio.hook
+++ b/src/zfs-utils/zfs-utils.initcpio.hook
@@ -103,12 +103,17 @@ zfs_mount_handler () {
     done
 }
 
-run_hook() {
+set_flags() {
     # Force import the pools, useful if the pool has not properly been exported using 'zpool export <pool>'
     [ ! "${zfs_force}" = "" ] && ZPOOL_FORCE="-f"
 
     # Add import directory to import command flags
     [ ! "${zfs_import_dir}" = "" ] && ZPOOL_IMPORT_FLAGS="${ZPOOL_IMPORT_FLAGS} -d ${zfs_import_dir}"
+    [ "${zfs_import_dir}" = "" ] && ZPOOL_IMPORT_FLAGS="${ZPOOL_IMPORT_FLAGS} -c /etc/zfs/zpool.cache"
+}
+
+run_hook() {
+    set_flags
 
     # Wait 15 seconds for ZFS devices to show up
     [ "${zfs_wait}" = "" ] && ZFS_WAIT="15" || ZFS_WAIT="${zfs_wait}"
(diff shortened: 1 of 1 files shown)


demizer:
Nice work! I want to test this before merging, but it looks good. Thank you.

minextu:
If anyone wants to help testing, I added this commit to my auto-build repo (https://github.com/minextu/archzfs)
```
[archzfs-ci]
Server = https://arch.et.tc/$repo/$arch
```
Key-ID: `45E7DCDCA0851D1B2B9564A009CCCF3EF41FCBCE`

lnicola:
@minextu would you recommend using that repository?

minextu:
@lnicola It's completely automated without testing the resulting packages. But I haven't had any issues so far. 
When I finally get automated testing done, we will probably have the same system on the normal archzfs repo.

lnicola:
@minextu Yeah, I installed your packages and they didn't eat my data :).

> But I haven't had any issues so far.

If you're dog-fooding them, it's fine.

minextu:
I merged it, thanks! This change will be included only in the git packages for now, to allow further testing. When zfs 0.7.10 get's released it'll be included there, too.

PR #240, opened 2018-04-09
Prefix the git revision with a 'g'

minextu:
All git packages are missing a 'g' prefix in versioning. 
This was caused by my commit in https://github.com/archzfs/archzfs/pull/200


minextu:
Related to #198

PR #261, opened 2018-07-12
Adding remote root filesystem decryption initcpio hook

ArnCo:
Here's a patch to allow the init process to continue if an encrypted root partition is unlocked via SSH on boot. 
Once the unlock via SSH is done, the "zfs load-key" process needs to be killed. Then, the condition [ "$(zfs get -H -o value keystatus "${encryptionroot}")" != "available" ] becomes false and breaks the loop.
I'll develop an external initcpio hook to implement remote unlocking via this repo: https://github.com/grazzolini/mkinitcpio-utils as it already implements a similar system for LUKS.

### DIFF
--- a/src/zfs-utils/PKGBUILD.sh
+++ b/src/zfs-utils/PKGBUILD.sh
@@ -64,6 +64,7 @@ package() {
     # Install the support files
     install -D -m644 "\${srcdir}"/zfs-utils.initcpio.hook "\${pkgdir}"/usr/lib/initcpio/hooks/zfs
     install -D -m644 "\${srcdir}"/zfs-utils.initcpio.install "\${pkgdir}"/usr/lib/initcpio/install/zfs
+    install -D -m644 "\${srcdir}"/zfs-utils.initcpio.zfsencryptssh.install "\${pkgdir}"/usr/lib/initcpio/install/zfsencryptssh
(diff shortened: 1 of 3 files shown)


ArnCo:
I opened an issue at grazzolini/mkinitcpio-utils, but I'm not sure if the maintainer is still active. Could it be a possibility to integrate the hook in the archzfs repo @minextu  ?

minextu:
That would be an option. Though I won't have time to test/merge this week.

ArnCo:
There's no hurry :) . I won't have time to submit it this week neither... I'll keep you updated.

ArnCo:
Hey @minextu , I just pushed the changes. Let me know if I can do something more to help :)

Cheers, ArnCo

minextu:
Awesome thanks! I changed it a bit to work without bootfs and without requiring Bash. It will be included when I next update packages.

minextu:
Commits I added:
106f3de2e1ad0858e9a542074c9916975a53d1fa
715074ffbc67bc0086f7c6aadf99ea43936d803f

ArnCo:
That's awesome, clever way to get rid of the bootfs flag, and well done for the hook signature.

Keep up the great work :)

ArnCo

sotiris-bos:
Can this be used with mkinitcpio-systemd-tool?

Is there any documentation on how to use it?

bglod:
@sotiris-bos I used this, worked well enough:

https://kernelpanic.cordier.work/en/blog/archlinux-zfs-remote-encrypted-root

PR #276, opened 2018-11-14
Initcpio hook: skip dataset mount with canmount=off property

ivan3182:
Datasets with canmount=off property should not be mounted


minextu:
Merged with 4001747c50cc416abbf2364df37ec3c2dc2da4b7. Thanks!

PR #363, opened 2020-10-18
Fix redirection in initcpio hook

kstolp:
Currently, the redirection is creating a file (`1`) instead of properly redirecting the output.


minextu:
Good catch, thanks for fixing this!

Issue #365, opened 2020-10-22
Allow users to skip zfs import in the late hook


Bronek:
Sometimes users do not want to import pools in the late hook and e.g. instead rely on `zfs-import-cache.service`.

Currently late hook has this check, which under some circumstances will skip `zfs import`:
```
run_latehook () {
    set_flags
    # only run zpool import, if flags were set (cache file found / zfs_import_dir specified)
    [ ! "${ZPOOL_IMPORT_FLAGS}" = "" ] && zpool import ${ZPOOL_IMPORT_FLAGS} -N -a ${ZPOOL_FORCE}
}
```

However, if the user wanted to use flags `zfs_import_dir` or `zfs_force` when importing the boot pool, then per above code we will always execute `zfs import` on all other pools as well. The same applies if there are no flags set in the kernel command line, however `zfs_mount_handler ()` decided to import boot pool read-only and set `ZPOOL_IMPORT_FLAGS=" -o readonly=on"` as a result.

It would be desirable to provide an option to skip the late hook `zfs import`, even if import flags are set.

Bronek:
The PR #366 presents one way to implement this.

minextu:
PR has been merged and a new version has been pushed. `zfs_boot_only=1` can now be set.
Thx @Bronek!

PR #366, opened 2020-10-22
Disable late hook if zfs_boot_only is set

Bronek:
This allows the user to rely on zfs-import-cache.service to import ZFS pools, if they want to.

This solves #365


minextu (approved):
Looks good. Thanks!

PR #367, opened 2020-10-25
Don't mount datasets with canmount=noauto

keidax:
I'm using ZFS as my root FS, with some (encrypted) datasets that I don't want mounted at boot time. I would like to boot the system without having to enter the password for the encrypted datasets.

This change will allow some datasets to be ignored at boot and mounted explicitly later, by respecting the `canmount=noauto` property.

My usecase is inspired by [this setup](https://talldanestale.dk/2020/04/06/zfs-and-homedir-encryption/).


minextu (approved):
Awesome, thx!

PR #369, opened 2020-11-02
Mount ZFS_DATASET regardless of canmount=noauto

rpthms:
The recent change in #367 caused my root dataset to not get mounted
because it had canmount set to noauto.

The root dataset should be mounted even if canmount is set to noauto.
This is because there can be multiple root datasets if one is using
boot environments with a tool like zectl (https://github.com/johnramsden/zectl).
The root datasets always have the same mountpoint (/), so you cannot
set canmount=on on all of them as that would result in overlapping mounts.


minextu (approved):
Sorry about breaking your setup. Never heard about zectl before.
Thanks for reporting and fixing!

PR #389, opened 2021-03-09
Support overriding encryption keylocation to prompt in initcpio hook

ahesford:
If an encrypted ZFS dataset specifies a file:// keylocation, relying on the default `zfs load-key` behavior will prevent the user from typing a passphrase even if its keyformat is passphrase. This renders an initcpio that does not contain the key file unbootable unless SSH has been preconfigured and the user connects remotely do unlock the dataset.

To allow some recovery, the hook now will attempt to force a prompt if the default load-key behavior failed to unlock the dataset and the keyformat is passphrase. This is still compatible with providing a key passphrase on, e.g., some USB device because the user can insert the device after the fact and then just interrupt the load-key prompt (or enter a garbage password three times) to retry the default load.

Also, zfs_decrypt_fs has been reworked to monitor the return value of ZFS operations, which allows graceful failure for ZFS versions that do not support encryption without having to scrape the help text of zfs.

### DIFF
--- a/src/zfs-utils/zfs-utils.initcpio.hook
+++ b/src/zfs-utils/zfs-utils.initcpio.hook
@@ -29,24 +29,44 @@ zfs_get_bootfs () {
 zfs_decrypt_fs() {
     dataset=$1
 
-    # check if 'zfs load-key' is available
-    zfs 2>&1 | grep load-key > /dev/null || return 0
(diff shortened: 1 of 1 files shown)


ahesford:
Ping

minextu (approved):
Thank you for this! Sorry for the delay in getting this merged

PR #396, opened 2021-04-28
Append user information to passwd

iamGavinJ:
make_etc_passwd() is overwriting user information `passwd` from earlier hooks.

Signed-off-by: Gavin Scurr <gavin@scurr.me>
Closes #N/A
Issue #N/A


minextu (approved):
Thank you!

PR #427, opened 2021-12-20
Wait for udev to settle if keylocation is a local file

### LINKED ISSUE #426: zfs_decrypt_fs should wait for udev to settle if keylocation is a local file
See https://github.com/openzfs/zfs/issues/12065.

puzzlemoondev:
If an encrypted dataset has a file keylocation on an external device, like an USB flash drive, zfs_decrypt_fs will fail to read the file.
A solution is to wait for udev events to be handled and keylocation to appear, which has been implemented in zfs-dracut (https://github.com/openzfs/zfs/blob/3fa5266d727a0e9506dfeed0118cbe7dd3c2a18d/contrib/dracut/90zfs/zfs-load-key.sh.in#L56).
I've adopted the solution here.
Closes #426

### DIFF
--- a/src/zfs-utils/zfs-utils.initcpio.hook
+++ b/src/zfs-utils/zfs-utils.initcpio.hook
@@ -58,18 +58,22 @@ zfs_decrypt_fs() {
             keyfile="${keylocation#file://}"
 
             # If file not yet exist, wait for udev to create device nodes
-            [ -r "${keyfile}"] || udevadm settle
+            if [ ! -r "${keyfile}" ]; then
+                udevadm settle
 
-            # Wait for udev up to 10 seconds
-            [ -r "${keyfile}" ] || {
-                echo "Waiting for key ${keyfile} for ${encryptionroot}..."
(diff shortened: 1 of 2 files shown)


src/zfs-utils/zfs-utils.initcpio.hook
```diff
@@ -44,14 +44,33 @@ zfs_decrypt_fs() {
     # Export encryption root to be used by other hooks (SSH)
     echo "${encryptionroot}" > /.encryptionroot
 
-    # If key location is a file, determine if it can by overridden by prompt
     prompt_override=""
-    if keylocation="$(zfs get -H -o value keylocation "${dataset}")"; then
+    if keylocation="$(zfs get -H -o value keylocation "${encryptionroot}")"; then
```
puzzlemoondev, line 48: Get keylocation value from encryption root because dataset keylocation might be `none`.

minextu:
Thank you! Sorry for the delay in merging this. The change will be in the testing environment once the build finishes

PR #447, opened 2022-05-18
Add /etc/fstab to initramfs

luc65r:
I spent way too much time figuring why initramfs wasn't mounting the root dataset.
My root dataset mountpoint is set to legacy. The hook was assuming the root partition was already mounted on `/new_root` and trying to read `/new_root/etc/fstab` to mount legacy datasets.

This is small fix that enables to mount the root dataset with mountpoint=legacy.


minextu:
Thank you and sorry for the late merge

Issue #454, opened 2022-07-20
zfs-utils: zed wrongly creates pidfiles in /usr/var/run 


leonghui:
Found the following files on all my Arch systems using `lostfiles`. They should be in `/var/run` instead.
```
[leonghui@me run]$ ls -ld $PWD/*
-rw-r--r-- 1 root root  4 Jul 20 22:46 /usr/var/run/zed.pid
-rw-r--r-- 1 root root 24 Jul 20 23:35 /usr/var/run/zed.state
```

FYI it was reported for the AUR zfs-utils package [too](https://aur.archlinux.org/packages/zfs-utils#comment-855943), but it was not fixed.

PR #455, opened 2022-07-20
zfs-utils: Fix wrong zed files in /usr/var/run

leonghui:
Fixes archzfs/archzfs#454

Before:
```
[leonghui@me run]$ ls -ld $PWD/*
-rw-r--r-- 1 root root  4 Jul 20 22:46 /usr/var/run/zed.pid
-rw-r--r-- 1 root root 24 Jul 20 23:35 /usr/var/run/zed.state
```

After:
```
[leonghui@me /]$ ls -ld $PWD/* /var/run | grep zed
-rw-r--r--  1 root root        7 Jul 20 23:56 /var/run/zed.pid
-rw-r--r--  1 root root       24 Jul 20 23:49 /var/run/zed.state
```


minextu:
Thanks!

PR #457, opened 2022-08-19
Fix unbootable system with `root=zfs` kernel parameter

haozhangphd:
If the kernel parameter is set to `root=zfs`, the system is unbootable because the `pool` variable is not set in the initcpio hook. The error message is as follows:
```
cannot import '': no such pool available
ERROR: ZFS: Unable to import pool .
```
This PR fixes this issue by making `root=zfs` behaves the same as `zfs=auto` in the kernel parameter, as implied by the help message.


minextu:
Thanks!

Issue #464, opened 2022-11-05
zfs-utils and mkinitcpio hook fails due to libcrypto.so.1.1 link


frebib:
OpenSSL 3.0 just dropped: https://archlinux.org/packages/core/x86_64/openssl/

```
(12/21) Install DKMS modules
==> dkms install --no-depmod nvidia/520.56.06 -k 5.15.77-1-lts
==> dkms install --no-depmod v4l2loopback/0.12.7 -k 5.15.77-1-lts
==> dkms install --no-depmod v4l2loopback/0.12.7 -k 6.0.7-zen1-1-zen
==> dkms install --no-depmod zfs/git -k 5.15.77-1-lts
==> dkms install --no-depmod nvidia/520.56.06 -k 6.0.7-zen1-1-zen
==> dkms install --no-depmod zfs/git -k 6.0.7-zen1-1-zen
==> depmod 5.15.77-1-lts
==> depmod 6.0.7-zen1-1-zen
(13/21) Signing kernel for secure boot
Signing Unsigned original image
Signing Unsigned original image
==> Building image from preset: /etc/mkinitcpio.d/linux-lts.preset: 'default'
  -> -k /boot/vmlinuz-linux-lts -c /etc/mkinitcpio.conf -g /boot/initramfs-linux-lts.img
==> Starting build: 5.15.77-1-lts
  -> Running build hook: [base]
  -> Running build hook: [udev]
  -> Running build hook: [autodetect]
  -> Running build hook: [keymap]
  -> Running build hook: [keyboard]
==> WARNING: Possibly missing firmware for module: xhci_pci
  -> Running build hook: [modconf]
  -> Running build hook: [block]
  -> Running build hook: [zfs]
==> ERROR: binary dependency `libcrypto.so.1.1' not found for `mount.zfs'
==> ERROR: binary dependency `libcrypto.so.1.1' not found for `zed'
==> ERROR: binary dependency `libcrypto.so.1.1' not found for `zfs'
==> ERROR: binary dependency `libcrypto.so.1.1' not found for `zinject'
==> ERROR: binary dependency `libcrypto.so.1.1' not found for `zpool'
==> ERROR: binary dependency `libcrypto.so.1.1' not found for `zstreamdump'
  -> Running build hook: [filesystems]
```

Workaround for now is to install the `openssl-1.1` package

frebib:
Rebuilding the https://github.com/archzfs/zfs-utils package seems to fix it.
[...]

PR #465, opened 2022-11-05
Bump utils pkgrel for openssl 3.0 rebuild

frebib:
zfs utility binaries link against libcrypto.so.1.1 in the existing package. Bump the pkgrel to force a rebuild against libcrypto.so.3

Fixes: https://github.com/archzfs/archzfs/issues/464
Signed-off-by: Joe Groocock <me@frebib.net>

### DIFF
--- a/src/kernels/_utils.sh
+++ b/src/kernels/_utils.sh
@@ -3,10 +3,10 @@ mode_name="utils"
 mode_desc="Select and use the utils packages"
 
 # version
-pkgrel="1"
+pkgrel="2"
 
 # Version for GIT packages
-pkgrel_git="1"
+pkgrel_git="2"
 zfs_git_commit=""
 zfs_git_url="https://github.com/zfsonlinux/zfs.git"
 


minextu:
Thanks!

bountin:
I'm on Manjaro and curious because the update did in fact break the initramfs building with an error that `libcrypto.so.3` is missing (AFAIR, I don't want to restore again). Manjaro doesn't have OpenSSL 3 in stable but only in unstable at the moment.

How can it be that a rebuild without apparent code changes break? [Also I'm aware that AUR is targeted for Arch and not other distros. Just curious how this OpenSSL 3 dependency sneaked in.]

minextu:
During a build, existing libraries are checked and linked against (on the system that does the compiling). So If you rebuild on manjaro it would link against the older openssl, same on an outdated arch install for that matter.

bountin:
I remembered today, that I can simply try again and backup the initramfs files before building :thinking: And found out that installing the old package also works.

And I've also found that I'm stupid because of course `zfs-utils` is the built binary linked to OpenSSL 3 and `zfs-utils-git` would probably just work ... Sorry again for hijacking this issue and thanks for the help!

Just adding the error output here , so that Google can point other confused souls to this answer:
```
(4/6) Updating linux initcpios...
==> Building image from preset: /etc/mkinitcpio.d/linux515.preset: 'default'
  -> -k /boot/vmlinuz-5.15-x86_64 -c /etc/mkinitcpio.conf -g /boot/initramfs-5.15-x86_64.img
==> Starting build: 5.15.76-1-MANJARO
  -> Running build hook: [base]
  -> Running build hook: [udev]
  -> Running build hook: [autodetect]
  -> Running build hook: [modconf]
  -> Running build hook: [block]
  -> Running build hook: [keyboard]
  -> Running build hook: [zfs]
==> ERROR: binary dependency `libcrypto.so.3' not found for `mount.zfs'
==> ERROR: binary dependency `libcrypto.so.3' not found for `zed'
==> ERROR: binary dependency `libcrypto.so.3' not found for `zfs'
==> ERROR: binary dependency `libcrypto.so.3' not found for `zinject'
==> ERROR: binary dependency `libcrypto.so.3' not found for `zpool'
==> ERROR: binary dependency `libcrypto.so.3' not found for `zstreamdump'
```

R8s6:
Oddly, i did a full system upgrade on 2022-11-16 (with last update being in July 2022), I got the `libcrypto.so.3` not found error just like @bountin , and after disabling archzfs from `/etc/pacman.conf` and re-compiling zfs-utils, regenerating initramfs and grubs with seemingly no more errors, i still couldn't reboot back into system, with the error being `filesystem 'ZFS=zroot/ROOT/default' cannot be mounted`

Luckily i did a snapshot before upgrading so i'm back to normal but I gotta dig deeper to find out what went wrong. Will come back and update on this.

PR #542, opened 2024-08-07
Update to zfs version 2.2.5

SoongNoonien:
Hi!
I've updated to 2.2.5 but the issue #538 kind of remains.

PR #546, opened 2024-08-18
Fix package build issue caused by kmod-33

Bronek:
This fixes build error which is blocking #542 - the build error is unrelated to zfs 2.2.5 but instead it's a timing accident with kmod-33 release, as explained in the commit. I am not entirely sure that disabling `DEPMOD` is the right way to approach this, but that's the suggested solution in https://bbs.archlinux.org/viewtopic.php?pid=2190572#p2190572 and also what I tested.

The error looks like this
```
  DEPMOD  /build/zfs-linux-lts/pkg/zfs-linux-lts/usr/lib/modules/6.6.46-1-lts
depmod: ERROR: could not open directory /build/zfs-linux-lts/pkg/zfs-linux-lts/usr/usr/lib/modules/6.6.46-1-lts: No such file or directory
depmod: FATAL: could not search modules: No such file or directory
```

The commit which "broke" kmod for us is here

https://gitlab.archlinux.org/archlinux/packaging/packages/kmod/-/commit/0efd732cb78bc0b7851a8367f4dc8e6933f5b99d

This change was only released on 15th Aug in kmod-33 (almost half year after this change).

In another commit, this PR also fixes an unrelated innocuous/dummy error, which looks like this:
```
/bin/sh: line 1: /build/zfs-linux-lts/pkg/zfs-linux-lts-headers/usr/share/zfs/common.sh: No such file or directory
make[3]: [Makefile:13923: scripts-install-data-hook] Error 1 (ignored)
```

### DIFF
--- a/src/zfs/PKGBUILD.sh
+++ b/src/zfs/PKGBUILD.sh
@@ -21,6 +21,11 @@ sha256sums=("${zfs_src_hash}")
 license=("CDDL")
 depends=("kmod" "${zfs_utils_pkgname}" ${linux_depends})
 
+prepare() {
+    cd "${zfs_workdir}"
+    sed -ri 's|(KERNELRELEASE=@LINUX_VERSION@)|\1 DEPMOD=/doesnt/exist|' module/Makefile.in
+}
+
 build() {
     cd "${zfs_workdir}"
     ./autogen.sh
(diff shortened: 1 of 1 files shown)


n0xena:
works great - build works smoothly again

Bronek:
I think this is no longer needed, the above mentioned breaking change in `kmod` package has been reverted on Aug 22 and released with kmod-33-2

https://gitlab.archlinux.org/archlinux/packaging/packages/kmod/-/commit/5c47a72852d7dc90d0f378e2b174be7e33a54abe

PR #551, opened 2024-10-04
Add GitHub Actions

Doridian:
This is an (early) draft of how integration with GitHub Actions might look like.

I was bored today so I just did this, so no worries if another route is chosen entirely.

This draft also includes the update to 2.2.6 just to make CI pass (as in, build with all kernels)

TODO:
- [x] Actually publish a release from `master` and create repo files
- [x] Sign packages with GPG key from secrets
- [x] ~Investigate ways to actually figure out which packages to build~
- [x] Periodic rebuild (for new kernels etc)

Progress can be viewed here: https://github.com/Doridian/archzfs/actions

This expects an "Environment" named `Release`, with a `GPG_KEY_ID` variable and a `GPG_KEY_DATA` secret (the GPG key to sign the packages with).
If not present, it will instead publish an unsigned repo as the `latest` release.

It will also automatically keep the `latest` tag up to date with the `master` branch, to make sure the GitHub Release points at the latest commit.

Release can be seen here: https://github.com/Doridian/archzfs/releases/tag/latest

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,25 @@
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    paths:
+      - '**.sh'
+      - '.github/workflows/**'
+      - 'build-container/**'
+      - 'packages/**'
+name: Build
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4.2.1
+      - name: Build builder container
+        run: docker build -t archzfs-builder build-container
+      - name: Run builder container
+        run: docker run --privileged --rm -v "$(pwd):/src" archzfs-builder
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,52 @@
+on:
+  push:
+    branches: [master]
+    paths:
+      - '**.sh'
+      - '.github/workflows/**'
+      - 'build-container/**'
+      - 'packages/**'
+  schedule:
+    - cron: "4 2 * * *"
+
+name: Release
+
+concurrency:
+  group: release
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    environment: Release
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4.2.1
+      - name: Build builder container
+        run: docker build -t archzfs-builder build-container
+      - name: Run builder container
+        env:
+          GPG_KEY_DATA: "${{ secrets.GPG_KEY_DATA }}"
+          GPG_KEY_ID: "${{ vars.GPG_KEY_ID }}"
+        run: docker run -e GPG_KEY_DATA -e GPG_KEY_ID --privileged --rm -v "$(pwd):/src" archzfs-builder
+      - name: Release mainline
+        uses: ncipollo/release-action@v1.14.0
+        with:
+          name: experimental
+          tag: experimental
+          commit: ${{ github.sha }}
+          artifacts: ./repo/*
+          allowUpdates: true
+          artifactErrorsFailBuild: true
+          omitBody: true
+          omitBodyDuringUpdate: true
+          removeArtifacts: true
+      - uses: rickstaa/action-create-tag@v1.7.2
+        with:
+          tag: experimental
+          force_push_tag: true
--- a/build-container/Dockerfile
+++ b/build-container/Dockerfile
@@ -1,0 +1,40 @@
+FROM archlinux:base-devel
+
+# The following is from https://github.com/archzfs/archzfs-ci/blob/master/worker/Dockerfile
+RUN pacman -Syu --noconfirm --needed python-pipx python-twisted python-future git wget systemd-sysvcompat openresolv vi
+
+# add buildbot user and give passwordless sudo access (needed for archzfs build scripts)
+RUN groupadd -r buildbot && \
+    useradd -m -g buildbot buildbot && \
+    mkdir /worker && \
+    chown buildbot:buildbot /worker && \
+    echo "buildbot ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
+# aur prep
+RUN useradd aur && \
+    echo "aur ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
+# clean-chroot-manager
+USER aur
+RUN mkdir -p /tmp/ccm-install            && \
+    cd /tmp/ccm-install                  && \
+    wget https://aur.archlinux.org/cgit/aur.git/snapshot/clean-chroot-manager.tar.gz && \
+    tar -xvf clean-chroot-manager.tar.gz && \
+    cd clean-chroot-manager              && \
+    makepkg -si --noconfirm              && \
+    cd /tmp                              && \
+    rm -rfv /tmp/ccm-install
+
+USER root
+COPY ccm.conf /home/buildbot/.config/clean-chroot-manager.conf
+RUN ccm64 p && \
+    chown -R buildbot:buildbot /home/buildbot && \
+    mkdir -p /scratch/.buildroot
+
+COPY entrypoint.sh /entrypoint.sh
+COPY systemd-nspawn-wrapper /usr/local/bin/systemd-nspawn
+RUN systemd-machine-id-setup
+
+USER buildbot
+VOLUME /src
+ENTRYPOINT [ "/entrypoint.sh" ]
--- a/build-container/ccm.conf
+++ b/build-container/ccm.conf
@@ -1,0 +1,71 @@
+# Fully qualified path for build root.
+# This should not use a variable like $HOME.  If your machine has lots
+# of memory, consider locating this to tmpfs to avoid usage to the disk and
+# to minimize access times but know that unless you copy the contents to
+# physical media, it will not survive a reboot. See the manpage for tips.
+CHROOTPATH64="/scratch/.buildroot"
+
+# Number of threads makepkg in the clean chroot will use when building.
+# The typical rule is physical cores + 1.
+THREADS=9
+
+# Optionally uncomment and define a custom pacman.conf and/or a custom
+# makepkg.conf for the buildroot using a fully qualified path below.
+# Leaving these two undefined to use the system files.
+
+#CUSTOM_PACMAN_CONF='/usr/share/devtools/pacman.conf.d/extra-testing.conf'
+#CUSTOM_MAKEPKG_CONF='/usr/share/devtools/makepkg.conf.d/x86_64.conf'
+
+# Optionally uncomment and define a custom location and name for the local chroot
+# package repo.
+#REPO="/src/repo"
+REPO_NAME="chroot_local"
+
+# Optionally uncomment to pass the --nocheck flag to the build which will skip
+# the check function in the PKGBUILD is it is present.
+#NOCHECK=1
+
+# Optionally define the format of compression for compiled packages. Leave this
+# undefined to use the Arch default.
+PKGEXT=
+
+# If set, the value defined will be used in the buildroot's packages.
+PACKAGER="ArchZFS Project (https://github.com/archzfs)"
+
+# Set this variable to anything if you want to run namcap on the built package.
+RUNNAMCAP=
+
+# Set this to anything if you want makepkg to build using distcc for faster
+# compilation. You must have distcc nodes properly configured on volunteers
+# you define below.  It does NOT need to be running on the native environment.
+#
+# Alternatively, you can invoke ccm with the 'cd' flag to create the chroot
+# with distcc enabled rather than editing this value.
+#
+# For more on distcc, see: https://wiki.archlinux.org/index.php/Distcc
+RUNDISTCC=
+
+# This is only needed for users planning to build with distcc. Take care to
+# include the localhost plus all volunteers you define below. As a rule of thumb,
+# set to about twice the total number of available server CPUs. See the distcc
+# man page for more info.
+DISTCC_THREADS=
+
+# Define all machines in the distcc cluster below using the distcc syntax of:
+# "hostname/threads" and remember to list localhost/n first, followed by your
+# volunteer nodes listed in decreasing order of CPU power. Additional supported
+# options are passed through, see the manpage for distcc.
+#
+# In my experience, one sees best results using twice the number of physical
+# cores on the volunteer machines. In the example below foo is a quad and bar
+# is a dual.
+#DISTCC_HOSTS="localhost/9 foo/8 bar/4"
+
+# To build with ccache in the buildroot, uncomment and define the directory where
+# ccache will store its data below.  If you're using a custom makepkg.conf
+# (see below), you MUST enable ccache that file in addition to defining the path
+# below. For more info about ccache, see: https://wiki.archlinux.org/index.php/ccache
+#
+#CCACHE_DIR="/scratch/.ccache"
+
+CUSTOM_PACMAN_CONF="/usr/share/devtools/pacman.conf.d/extra.conf"
--- a/build-container/entrypoint.sh
+++ b/build-container/entrypoint.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+set -e
+
+if [ ! -z "${GPG_KEY_DATA-}" ]; then
+    if [ -z "${GPG_KEY_ID-}" ]; then
+        echo 'GPG_KEY_ID is not set, but GPG_KEY_DATA is set. Please set GPG_KEY_ID to the key ID of the key.'
+        exit 1
+    fi
+    gpg --import /dev/stdin <<<"${GPG_KEY_DATA}"
+fi
+
+# Only set -x here so we can't accidently print the GPG key up there
+set -x
+
+sudo chown -R buildbot:buildbot /src
+cd /src
+
+sed -i "/^THREADS=/s/9/$(nproc)/" ~/.config/clean-chroot-manager.conf
+sudo ccm64 d || true
+
+sudo bash build.sh -d -u all update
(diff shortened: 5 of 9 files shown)


Doridian:
This is now pretty much ready with the basics of publishing a repo.

There is no logic for periodic rebuilds or smart rebuilds, but modified the `release` action to have a cron trigger should be a very simple change, and with the builds taking 45 minutes even for a full rebuild, I am not sure if we need to bother being smarter about it with nightly builds?

kerberizer:
@Doridian Thank you so much for taking on this task! I really appreciate the work you've done so far—it looks great. I've sent you an invite in case you'd prefer to continue working directly on this branch here. It might be more convenient for running the CI and could also simplify future maintenance.

Doridian:
> @Doridian Thank you so much for taking on this task! I really appreciate the work you've done so far—it looks great. I've sent you an invite in case you'd prefer to continue working directly on this branch here. It might be more convenient for running the CI and could also simplify future maintenance.

Thanks for the invite. Right now it is actually easier to work on my fork, until this is merged in. (Because that means I can screw up the `master` branch without any issues), but this will definitely be easier to work with for future additions :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: docker run -e GPG_KEY_DATA -e GPG_KEY_ID --privileged --rm -v "$(pwd):/src" archzfs-builder
+      - name: Release mainline
+        uses: ncipollo/release-action@v1
+        with:
+          tag: latest
+          name: latest
+          commit: ${{ github.sha }}
+          artifacts: ./repo/*
+          removeArtifacts: true
+          replacesArtifacts: true
```
kerberizer, line 42: This seems to be the default.

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+          allowUpdates: true
+          omitBody: true
+          omitBodyDuringUpdate: true
+          artifactErrorsFailBuild: true
```
kerberizer, line 46: Let's sort those parameters.

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Run builder container
+        env:
+          GPG_KEY_DATA: "${{ secrets.GPG_KEY_DATA }}"
+          GPG_KEY_ID: "${{ vars.GPG_KEY_ID }}"
+        run: docker run -e GPG_KEY_DATA -e GPG_KEY_ID --privileged --rm -v "$(pwd):/src" archzfs-builder
+      - name: Release mainline
+        uses: ncipollo/release-action@v1
+        with:
+          tag: latest
+          name: latest
```
kerberizer, line 38: We must remember to change this if we decide on a different name for the repo, right?

build-container/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM archlinux:base-devel
+
+# The following is from https://github.com/archzfs/archzfs-ci/blob/master/worker/Dockerfile
+RUN pacman -Syu --noconfirm --needed python-pipx python-twisted python-future git wget systemd-sysvcompat openresolv vi
```
kerberizer, line 4: It would be nice to get rid of some of these dependencies at some point, but let's not create unnecessary work now. Excellent job with those comments, BTW! The are really helpful.

build-container/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# aur prep
+RUN useradd aur && \
+    echo "aur ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
+# clean-chroot-manager
+USER aur
+RUN mkdir -p /tmp/ccm-install            && \
+    cd /tmp/ccm-install                  && \
+    wget https://aur.archlinux.org/cgit/aur.git/snapshot/clean-chroot-manager.tar.gz && \
```
kerberizer, line 21: We could use `curl` here, especially if we get rid of `wget` at some point, but it's okay for now IMO. I suppose we could've used `git` too, but the simpler, the better, so fetching the tar.gz is a good call I think.

build-container/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    tar -xvf clean-chroot-manager.tar.gz && \
+    cd clean-chroot-manager              && \
+    makepkg -si --noconfirm              && \
+    cd /tmp                              && \
+    rm -rfv /tmp/ccm-install
+
+USER buildbot
```
kerberizer, line 28: Seeing that all commands below use `sudo`, couldn't we just switch to `root`?

build-container/ccm.conf
```diff
@@ -0,0 +1,71 @@
+# Fully qualified path for build root.
+# This should not use a variable like $HOME.  If your machine has lots
+# of memory, consider locating this to tmpfs to avoid usage to the disk and
+# to minimize access times but know that unless you copy the contents to
+# physical media, it will not survive a reboot. See the manpage for tips.
+CHROOTPATH64="/scratch/.buildroot"
+
+# Number of threads makepkg in the clean chroot will use when building.
+# The typical rule is physical cores + 1.
+THREADS=9
```
kerberizer, line 10: If this isn't overridden somewhere else, maybe we could adjust it with something like this from either `build`/`release.yml` or `entrypoint.sh`?
```sh
NUMPROC="$(awk '/cpu cores/ { print $4; exit}' /proc/cpuinfo)"
sed -i "/^THREADS=/s/9/${NUMPROC}/" /path/to/ccm.conf
```

build-container/ccm.conf
```diff
@@ -0,0 +1,71 @@
...
+# Optionally uncomment to pass the --nocheck flag to the build which will skip
+# the check function in the PKGBUILD is it is present.
+#NOCHECK=1
+
+# Optionally define the format of compression for compiled packages. Leave this
+# undefined to use the Arch default.
+PKGEXT=
+
+# If set, the value defined will be used in the buildroot's packages.
+PACKAGER=
```
kerberizer, line 33: Is this set somewhere else? Where do we really set the PACKAGER entry for the packages?

build-container/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+set -e
+
+if [ ! -z "${GPG_KEY_DATA-}" ]; then
+    if [ -z "${GPG_KEY_ID-}" ]; then
+        echo 'GPG_KEY_ID is not set, but GPG_KEY_DATA is set. Please set GPG_KEY_ID to the key ID of the key.'
+        exit 1
+    fi
+    echo "${GPG_KEY_DATA}" | gpg --import /dev/stdin
```
kerberizer, line 9: A slightly safer way to do it, which wouldn't expose the key even with `set -x`, could be
```sh
gpg --import <<<"${GPG_KEY_DATA}"
```

build-container/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+# Only set -x here so we can't accidently print the GPG key up there
+set -x
```
kerberizer, line 13: Good job with these comments (and foresight as well), indeed!

build-container/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+
+sudo chown -R buildbot:buildbot /src
+cd /src
+
+sed -i 's/jan/buildbot/' conf.sh && sudo ccm64 d || true
```
kerberizer, line 18: Why not change it in `conf.sh` outright?

mirror.sh
```diff
@@ -0,0 +1,43 @@
...
+TMPDIR='./repo-tmp'
+# End of config settings
+
+rm -rf "${TMPDIR}"
+mkdir -p "${OUTDIR}" "${TMPDIR}"
+
+# TODO: Update to use the upstream repository
+RELEASE_INFO="$(curl -s https://api.github.com/repos/Doridian/archzfs/releases/latest)"
+
+FILE_INFO="$(echo "${RELEASE_INFO}" | jq '.assets | map(.browser_download_url + "|" + .updated_at) | join("\n")' -r)"
```
kerberizer, line 16: I wonder if reading this in an array wouldn't be more foolproof, e.g.
```sh
readarray -t FILE_INFO < <(echo "${RELEASE_INFO}" | jq '.assets | map(.browser_download_url + "|" + .updated_at) | join("\n")' -r)
```
and iterate below with
```sh
for info in "${FILE_INFO[@]}"; do
```

mirror.sh
```diff
@@ -0,0 +1,43 @@
...
+
+for info in $FILE_INFO; do
+    url="$(echo "${info}" | cut -d'|' -f1)"
```
kerberizer, line 19: this could perhaps be simplified to
```sh
url="${info%%|*}"
```
or, since we don't expect more than two fields
```sh
url="${info%|*}"
```

mirror.sh
```diff
@@ -0,0 +1,43 @@
...
+    tmp_filename="${TMPDIR}/$(basename "${url}")"
+    filename="${OUTDIR}/$(basename "${url}")"
+
+    updated_at_str="$(echo "${info}" | cut -d'|' -f2)"
```
kerberizer, line 23: same as above
```sh
updated_at_str="${info#*|}"
```

kerberizer (commented):
Overall, outstanding work, indeed. None of the inline comments are really critical, though it's probably important to optimize the CI runs with those path conditions.

build-container/entrypoint.sh
Doridian: Not sure, this is taken from archzfs-ci. We could just change the default, yeah

mirror.sh
Doridian: I haven't really used bash arrays at all, but yeah, that seems like a nicer solution, despite being a little more complex maybe? Yeah not sure, bash is never pretty to me

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    environment: Release
```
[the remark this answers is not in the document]
Doridian: Correct, yes.

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Build builder container
+        run: docker build -t archzfs-builder build-container
+      - name: Run builder container
+        env:
+          GPG_KEY_DATA: "${{ secrets.GPG_KEY_DATA }}"
+          GPG_KEY_ID: "${{ vars.GPG_KEY_ID }}"
```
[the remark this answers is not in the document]
Doridian: in my forked test setup, yeah. GPG_KEY_ID is an environment (GitHub term) variable and GPG_KEY_DATA is the ASCII armored GPG secret key as an environment secret

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
Doridian: Oh, turns out I completely misread how `fetch-depth` works. Yeah we need no history at all. I'll fix that.

build-container/entrypoint.sh
Doridian: Oh, I completely forgot about the `<<<`. Yeah that'd be neater for sure

build-container/ccm.conf
Doridian: This config is taken from the current archzfs-ci (since that only appends a line to the bottom of this default config).

Looking at the [current published archzfs packages](https://zxcvfdsa.com/archzfs/archzfs/x86_64/), those also have `Unknown Packager` in their metadata.

We could set it to something like "archzfs-ci" maybe? Not sure.

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
...
+  contents: read
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
Doridian: If we pin, I suggest we pin straight to git hashes, as dependabot / renovate can still upgrade these, but they also give us the added benefit of being unable to change.

Updated (at least via Renovate, haven't used Dependabot in a while) would look like this: https://github.com/FoxDenHome/LCDify/pull/20/files

build-container/Dockerfile
Doridian: I will try that, but I wonder if some tooling is dependant on being run via sudo (such that it can detect the low-privileged username etc).
We'll see :3

.github/workflows/build.yml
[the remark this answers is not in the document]
kerberizer: > but they also give us the added benefit of being unable to change.

Sorry, I'm afraid I didn't get this. The versions would have their own benefit of being easier to track for specific changes directly in the release notes, instead of having to seek for the matching commit hashes. I might be missing something of course.

mirror.sh
```diff
@@ -0,0 +1,43 @@
...
+# Config settings
+OUTDIR='./repo'
+TMPDIR='./repo-tmp'
+# End of config settings
+
+rm -rf "${TMPDIR}"
+mkdir -p "${OUTDIR}" "${TMPDIR}"
+
+# TODO: Update to use the upstream repository
+RELEASE_INFO="$(curl -s https://api.github.com/repos/Doridian/archzfs/releases/latest)"
```
kerberizer, line 14: We need to remember to update this before merging.

build-container/Dockerfile
kerberizer: It's just for the three commands in the RUN block below. Only `ccm64` might care, I guess, and most likely not. The final `USER buildbot` at the end of the file should stay, of course.

build-container/ccm.conf
kerberizer: > Looking at the [current published archzfs packages](https://zxcvfdsa.com/archzfs/archzfs/x86_64/), those also have `Unknown Packager` in their metadata.

Funny that it seems to have been like this all these years :)

> We could set it to something like "archzfs-ci" maybe? Not sure.

Supposedly it should be "name/email of the person or organization building packages" (from makepkg.conf), but I wonder if we could use instead, e.g. `ZFS on Arch Linux (https://github.com/archzfs)`. Namcap will likely dislike it though. Let's maybe put just `ZFS on Arch Linux` for now and revisit it after merging.

mirror.sh
kerberizer: Main advantage that comes to mind is that it would handle spaces in the entries gracefully. Not that we expect such spaces in our case, but maybe better safe than sorry. :D

.github/workflows/build.yml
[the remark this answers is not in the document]
Doridian: I was thinking since git hashes are immutable, but git tags are only suggested to be immutable. But yeah, probably not worth it to sacrifice readability

.github/workflows/release.yml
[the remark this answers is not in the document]
kerberizer: Done.

.github/workflows/release.yml
Doridian: Indeed, this and the below action to rename the tag

build-container/entrypoint.sh
```diff
@@ -15,6 +15,7 @@ set -x
 sudo chown -R buildbot:buildbot /src
 cd /src
 
+sed -i "/^THREADS=/s/9/$(nproc)/" ~/.config/clean-chroot-manager.conf
```
kerberizer, line 18: `nproc` is indeed simpler, but it gives the logical number of cores. Then again, maybe we're okay with this?

build-container/ccm.conf
kerberizer: Let's set this to `ArchZFS Project (https://github.com/archzfs)`, perhaps? It's what I set as name in the PGP key. Not ideal perhaps, but if needed we can revisit it later.

kerberizer:
I've set the GPG key things, so apart from a few non-critical nitpicks, probably only important thing that remains is to finalize the repo/release name (though we could change this later as well).

build-container/entrypoint.sh
Doridian: @kerberizer Usually the advice I heard (and the strategy I went with) is to set the number of parallel builds to logical cores (sometimes +1) to take as much advantage of any SMP as possible

Doridian:
> only important thing that remains is to finalize the repo/release name

Yeah, I think so, too.

build-container/entrypoint.sh
kerberizer: @Doridian Yup, Graysky's comment was "The typical rule is physical cores + 1", but this is indeed a classical case of "it depends". It will probably be best to test different values, but that's not a priority right now.

johnramsden:
This seems very solid. I also wonder if we could take advantage of the existence of the docker container and make it a supported way for people to do local builds (if they desire). I had some difficulty initially with clean-chroot, and making sure my environment was correct, this would make that process a lot easier. Once this gets merged I could add some instructions on the wiki.

kerberizer (approved):
@Doridian I think we can proceed with merging this. I suggest we go with the `experimental` name for the release for now. I've also uploaded the `0xABEC0A1208037BE9` key to the keyservers, so users can directly import it. Let's keep it this simple way for testing until we're ready with our own archzfs-keyring package (I'm working on it, but it'll take me some more time).

Once again, outstanding work. Thank you so much for it!

Doridian:
@kerberizer This is ready to merge whenever now. I changed the references etc :3

But we should merge https://github.com/archzfs/archzfs/pull/550 also since it is better to just use the pre-existing PR, and I'll remove the commit from my PR

kerberizer:
@Doridian good call, thanks! I've merged #550 and I'm leaving to you the pleasure to merge this one ;)

PR #556, opened 2024-11-04
ops: initial dependabot config

kerberizer:
I just realized that we don't have dependabot config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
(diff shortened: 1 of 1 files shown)


kerberizer:
We can add more assignees, of course.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+---
+version: 2
+updates:
+
```
Doridian, line 4: NIT: Blank line seems a bit odd here?

Doridian (approved):

johnramsden (approved):

Issue #560, opened 2024-11-25
Breakage of stable version of kernel prevents release of LTS version.


GregoryLand:
Right now the automatic builds are broken. https://github.com/archzfs/archzfs/actions/runs/12002432031

```
make  all-recursive
make[1]: Entering directory '/build/zfs-linux/src/zfs-2.2.6'
Making all in include
make[2]: Entering directory '/build/zfs-linux/src/zfs-2.2.6/include'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/build/zfs-linux/src/zfs-2.2.6/include'
Making all in module
make[2]: Entering directory '/build/zfs-linux/src/zfs-2.2.6/module'
mkdir -p os/linux/spl/
mkdir -p avl/ icp/ icp/algs/aes/ icp/algs/blake3/ icp/algs/edonr/ icp/algs/modes/ icp/algs/sha2/ icp/algs/skein/ icp/api/ icp/asm-aarch64/blake3/ icp/asm-aarch64/sha2/ icp/asm-arm/sha2/ icp/asm-ppc64/blake3/ icp/asm-ppc64/sha2/ icp/asm-x86_64/aes/ icp/asm-x86_64/blake3/ icp/asm-x86_64/modes/ icp/asm-x86_64/sha2/ icp/core/ icp/io/ icp/spi/ lua/ lua/setjmp/ nvpair/ os/linux/zfs/ unicode/ zcommon/ zfs/ zstd/ zstd/lib/common/ zstd/lib/compress/ zstd/lib/decompress/
make -C /usr/lib/modules/6.12.1-arch1-1/build  \
	  \
	M="$PWD"  CONFIG_ZFS=m modules
make[3]: Entering directory '/usr/lib/modules/6.12.1-arch1-1/build'
  CC [M]  /build/zfs-linux/src/zfs-2.2.6/module/os/linux/spl/spl-atomic.o
  CC [M]  /build/zfs-linux/src/zfs-2.2.6/module/os/linux/spl/spl-condvar.o
  CC [M]  /build/zfs-linux/src/zfs-2.2.6/module/os/linux/spl/spl-cred.o
  CC [M]  /build/zfs-linux/src/zfs-2.2.6/module/os/linux/spl/spl-err.o
  CC [M]  /build/zfs-linux/src/zfs-2.2.6/module/os/linux/spl/spl-generic.o
[...]

PR #561, opened 2024-11-26
Fix stable breaking other kernels

GregoryLand:
Stopped errors in build from exiting script.
Download last release repo to use as base for building new repo so we don't lose packages.

Problems:
Builds fail silently but this is a quick fix to get builds working again. https://github.com/archzfs/archzfs/issues/560

### DIFF
--- a/build-container/entrypoint.sh
+++ b/build-container/entrypoint.sh
@@ -24,6 +24,7 @@ build() {
     sudo bash build.sh -d "$1" make
 }
 
+set +e 
 build utils
 
 build std
@@ -31,6 +32,7 @@ build lts
 build hardened
 build zen
 build dkms
+set -e
 
 # Not implemented, yet, as documented in archzfs-ci
 # sudo bash test.sh ...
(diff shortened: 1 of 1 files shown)


Doridian:
@GregoryLand Your PR looks good mostly. There's only one concern: Right now this downloads everything from the release, not just failed builds. Where this will matter is when we increment versions (or the kernel does, as that's part of the package version of course)

We need to figure out a way to either:
- Prune artifacts we built successfully
- Only download artifacts we failed to build (this would allow us to fail a build if we both don't have old artifacts as well as can't build new ones the easiest, which would be a neat feature, but not required as it is a condition that can only happen on "startup"/bringup of a new repo)

Furthermore, we should probably allow this request to 404, in case we add a new release path (such as we will once we launch the `stable` repo, for example).

GregoryLand:
Aaaa right.  Okay I see the issue.  We would get into a situation were we are stacking up previous packages into the repo. I will try to poke that after work today and resolve it.

The more I look at it the more I think it makes sense to split the kernels into different builds somehow.  That will take significantly more work.

Doridian:
Closing this as we have merged a solution to this issue now.

PR #566, opened 2024-12-14
Handle individual failed kernel builds

Doridian:
This is my idea for a (fairly) simple solution for the "one or more of the kernels fails to build" issue.

It grabs the `.db` file, and then can use that to find out which packages are actually affected by a failed build and downloads exactly those when needed, thus avoiding the "doubling packages" issue.

`dkms` and `util` are not allowed to fail, as they don't depend on a kernel version.

Further, it also sets the `-s` flag (this is done in the same PR as otherwise I'd have to make `failover` able to handle those packages as well. And I do not think we should offer ZFS git packages (and from the discussions, it seems most agree), especially not on anything we eventually want to be a "stable" repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,9 @@ concurrency:
 permissions:
   contents: write
 
+env:
+  RELEASE_NAME: experimental
+
 jobs:
   release:
     name: Release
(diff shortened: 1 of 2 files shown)


Doridian:
@kerberizer @minextu Might be good to get an approval on this now that we're blocked on linux stable again (6.13 is out)

n0xena:
has someone else the permission to approve and merge this?

kerberizer:
Sorry, guys, had some difficult time, but I'll take a look at it asap

kerberizer:
@Doridian The release action unfortunately is failing. I'll see if I can spot what the problem might be.

kerberizer:
From very first look, it might be a mix between `$RELEASE_TYPE` and `$RELEASE_NAME`

kerberizer:
#575 seems to have fixed the building step in both build and release actions (no idea right now how the tests in this PR passed), but now the specific release step is failing with "permissions denied". Will continue investigating

PR #577, opened 2025-02-14
Fix release and increase failover resilience

Doridian:
This renames `build.yml` to `test.yml` and adds it pushing a dummy release to make sure the process works.


kerberizer:
Excellent idea to have releasing tested as well!

kerberizer:
```
:: Synchronizing package databases...
 chroot_local downloading...
 core downloading...
 extra downloading...
```
Perhaps it isn't seeing the additional repo as expected?

I'll see if I can find time to have a deeper look at it as well

kerberizer:
Could it be that the repo we're adding isn't added to the chroot that is used for the build?

.github/workflows/release.yml
```diff
@@ -36,7 +37,7 @@ jobs:
         env:
           GPG_KEY_DATA: "${{ secrets.GPG_KEY_DATA }}"
           GPG_KEY_ID: "${{ vars.GPG_KEY_ID }}"
-        run: docker run -e GPG_KEY_DATA -e GPG_KEY_ID -e RELEASE_NAME --privileged --rm -v "$(pwd):/src" archzfs-builder
+        run: docker run -e GPG_KEY_DATA -e GPG_KEY_ID -e FAILOVER_RELEASE_NAME -e RELEASE_NAME --privileged --rm -v "$(pwd):/src" archzfs-builder
```
kerberizer, line 40: It seems we can drop `-e RELEASE_NAME` with the latest changes?

.github/workflows/test.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: Test release
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4.2.1
+      - name: Build builder container
+        run: docker build -t archzfs-builder build-container
+      - name: Run builder container
+        run: docker run -e RELEASE_NAME -e FAILOVER_RELEASE_NAME --privileged --rm -v "$(pwd):/src" archzfs-builder
```
kerberizer, line 34: It seems we can drop `-e RELEASE_NAME` with the latest changes?

kerberizer (approved):
Outstanding work, thank you! Not sure only if you want to squash it before merging.

PR #578, opened 2025-02-14
Verify failover is properly signed

Doridian:
Just some additional insurance we can't ever accidentally failover to untrusted packages

### DIFF
--- a/build-container/entrypoint.sh
+++ b/build-container/entrypoint.sh
@@ -68,6 +68,8 @@ failover() {
             set -x
             tmp_file="$(mktemp)"
             curl -f -o "${tmp_file}" -L "${FAILOVER_BASE_URL}/${pkgfile}"
+            curl -f -o "${tmp_file}.sig" -L "${FAILOVER_BASE_URL}/${pkgfile}.sig"
+            gpg --verify "${tmp_file}.sig" "${tmp_file}"
             sudo mv "${tmp_file}" "/scratch/.buildroot/root/repo/${pkgfile}"
             set +x
         fi
--- a/build-container/entrypoint.sh
+++ b/build-container/entrypoint.sh
@@ -23,7 +23,7 @@ if [ ! -z "${FAILOVER_RELEASE_NAME}" ]; then
     db_file="archzfs.db.tar.xz"
     curl -f -o "${db_file}" -L "${FAILOVER_BASE_URL}/${db_file}"
     curl -f -o "${db_file}.sig" -L "${FAILOVER_BASE_URL}/${db_file}.sig"
-    if ! gpg --verify "${db_file}.sig" "${db_file}"; then
+    if ! gpg --trust-model direct --verify "${db_file}.sig" "${db_file}"; then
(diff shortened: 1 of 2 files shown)


build-container/entrypoint.sh
```diff
@@ -18,7 +20,14 @@ if [ ! -z "${FAILOVER_RELEASE_NAME}" ]; then
     FAILOVER_REPO_DIR="$(mktemp -d)"
     cd "${FAILOVER_REPO_DIR}"
     FAILOVER_BASE_URL="https://github.com/archzfs/archzfs/releases/download/${FAILOVER_RELEASE_NAME}"
-    if ! curl -fsL "${FAILOVER_BASE_URL}/archzfs.db.tar.xz" | tar xvJ; then
+    db_file="archzfs.db.tar.xz"
+    curl -f -o "${db_file}" -L "${FAILOVER_BASE_URL}/${db_file}"
+    curl -f -o "${db_file}.sig" -L "${FAILOVER_BASE_URL}/${db_file}.sig"
```
kerberizer, line 25: Perhaps we want to keep these curl executions in an `if !` like the next steps? Something like "Couldn't fetch failover..."?

build-container/entrypoint.sh
```diff
@@ -18,7 +20,14 @@ if [ ! -z "${FAILOVER_RELEASE_NAME}" ]; then
...
+    if ! gpg --verify "${db_file}.sig" "${db_file}"; then
+        echo 'Failover signature verification failed, failover impossible!'
+        FAILOVER_BASE_URL=""
+        FAILOVER_REPO_DIR=""
+    elif ! tar xvJf "${db_file}"; then
         echo 'Failover not found, failover impossible!'
```
kerberizer, line 31: With only `tar` being tested now, perhaps the message could be something like "Unreadable failover..."?

build-container/entrypoint.sh
```diff
@@ -68,6 +77,8 @@ failover() {
             set -x
             tmp_file="$(mktemp)"
             curl -f -o "${tmp_file}" -L "${FAILOVER_BASE_URL}/${pkgfile}"
+            curl -f -o "${tmp_file}.sig" -L "${FAILOVER_BASE_URL}/${pkgfile}.sig"
+            gpg --verify "${tmp_file}.sig" "${tmp_file}"
             sudo mv "${tmp_file}" "/scratch/.buildroot/root/repo/${pkgfile}"
```
kerberizer, line 82: I wonder if we don't want to first check that all files in the failover verify correctly and only then copy everything to the repo. This would prevent possible partially fetched failovers. On the other hand, I'm not sure if this would be meaningfully beneficial in practice.

kerberizer (approved):
Great idea!

build-container/entrypoint.sh
Doridian: @kerberizer We'd never end up with partial failovers as a single failing will fail the entire build, so the order would only change how early we abort a build, not whether we abort a build at all

PR #581, opened 2025-02-23
Fix mirror.sh: curl URL fixed and fail on errors

Doridian:
Setting up a local mirror I realized that the API URL no longer works.

So I updated that and also added `-f` to make it actually error and not fail silently


kerberizer (approved):
LGTM, thank you!

PR #583, opened 2025-03-03
add missing ./configure parameters to dkms

### LINKED ISSUE #579: different configure options for DKMS
this is practical a re-raise of an issue which came up on my fork: https://github.com/n0xena/archzfs/issues/17
while investigating I ended up getting the same as with my regular builds and why this repo doesn't provide 6.13 yet:
```
configure: error: 
        *** Cannot build against kernel version 6.13.3-arch1-1.
        *** The maximum supported kernel version is 6.12.
```
on my fork I use the --enable-linux-experimental as suggested zfs upstream: https://github.com/openzfs/zfs/discussions/16574#discussioncomment-11360884
https://github.com/n0xena/archzfs/blob/87f6b40a2f0d5dd1a703a654ac5bf092fa3748ed/src/zfs/PKGBUILD.sh#L30-L40
digging further down and executing DKMS manual got me this:
```
Autoinstall of module zfs/2.3.0 for kernel 6.13.3-arch1-1 (x86_64)
Running the pre_build script...(bad exit status: 1)
Failed command:
cd /var/lib/dkms/zfs/2.3.0/build/ && /var/lib/dkms/zfs/2.3.0/build/configure --disable-dependency-tracking --prefix=/usr --with-config=kernel --with-linux=/usr/lib/modules/6.13.3-arch1-1/build --with-linux-obj=/usr/lib/modules/6.13.3-arch1-1/build --with-qat=
```
so clearly the ./configure paramters vastly differs
back at some point when #555 was stil an issue before it was converted to a discusion I raised the question about the specific ./configure parameters to which @minextu replied: "because arch requires them this way"
I also looked up the arch wiki about dkms and pkgbuild - and there seems a way to give a config file for like custom ./configure and stuff

questions:
Why isn't this done for DKMS?
Doesn't this somewhat break arch dkms requirements, does it?
Should the ./configure parameters used for the regular builds be somehow injected into the dkms build?
Or was this not done on purpose?
Anyway - is there a way to smuggle in the experimental flag to get dkms build on non-standard 6.13 builds? If so: How?

I'm aware that the overall consent is: "Let's wait until zfs properly supports 6.13 with next update" - and according to what I read on zfs upstream they already work on support for 6.14 - so it can be assumed official support for 6.13 and maybe 6.14 is likely to come soon.

I and I guess https://github.com/alaricljs would appreciate if someone can give a helping hand here.

Thanks upfront for any help - appreciating your time.

n0xena:
closes #579 

as noted in 579: minextu once replied to my question about the ./configure paramters: "because that's what arch require" or something like this
so while figure out a way to inject --enable-linux-experimental to enable dkms build for 6.13 I also added the missing parameters from regular kernel PKGBUILDs
also bumping pkgrel version

### DIFF
--- a/src/kernels/dkms.sh
+++ b/src/kernels/dkms.sh
@@ -3,7 +3,7 @@ mode_name="dkms"
 mode_desc="Select and use the dkms packages"
 
 # version
-pkgrel="1"
+pkgrel="2"
 
 # Version for GIT packages
 pkgrel_git="1"
--- a/src/zfs-dkms/PKGBUILD.sh
+++ b/src/zfs-dkms/PKGBUILD.sh
@@ -10,8 +10,10 @@ pkgrel=${zfs_pkgrel}
 makedepends=(${zfs_makedepends})
 arch=("x86_64")
 url="https://openzfs.org/"
-source=("${zfs_src_target}")
-sha256sums=("${zfs_src_hash}")
+source=("${zfs_src_target}"
+    "dkms-configure.patch")
+sha256sums=("${zfs_src_hash}"
+    "cdb0121720197fe4c00f2725ea5a85de7be0f58806234a6c998de2e26b198334")
 license=("CDDL")
 depends=("${zfs_utils_pkgname}" "lsb-release" "dkms")
 provides=("zfs" "zfs-headers" "spl" "spl-headers")
@@ -27,6 +29,8 @@ build() {
 package() {
     dkmsdir="\${pkgdir}/usr/src/zfs-${zfs_mod_ver}"
     install -d "\${dkmsdir}"
+    cd ${zfs_workdir}
+    patch -Np1 -i ../dkms-configure.patch
     cp -a ${zfs_workdir}/. \${dkmsdir}
 
     cd "\${dkmsdir}"
--- a/src/zfs-dkms/dkms-configure.patch
+++ b/src/zfs-dkms/dkms-configure.patch
@@ -1,0 +1,17 @@
+diff -ura a/scripts/dkms.mkconf b/scripts/dkms.mkconf
+--- a/scripts/dkms.mkconf	2025-03-03 19:41:03.321389279 +0100
++++ b/scripts/dkms.mkconf	2025-03-03 19:58:45.609878706 +0100
+@@ -29,6 +29,13 @@
+   --disable-dependency-tracking
+   --prefix=/usr
+   --with-config=kernel
++  --sysconfdir=/etc
++  --sbindir=/usr/bin
++  --libdir=/usr/lib
++  --datadir=/usr/share
++  --includedir=/usr/include
++  --with-udevdir=/usr/lib/udev
++  --libexecdir=/usr/lib
+   --with-linux=\$(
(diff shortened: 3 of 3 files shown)


kerberizer:
Thank you!

This PR is really helpful if we do decide at some point to use the `--enable-linux-experimental` configure flag for the `experimental` repo.

Beyond that flag, however, I'm probably divided on it. It does indeed bring the `configure` options in sync with the non-DKMS packages. OTOH, it touches upstream and, while the change is relatively simple, this can always bring unforeseen consequences.

The main question here, I think, is what can we gain in practice from such change that could outweigh the risks of messing with upstream and complicating our own workflow.

I tested the build with and without the added `configure` options (the commands I used are below for reference), and the produced `spl.ko` and `zfs.ko` modules were binary identical.

```sh
git clone https://github.com/openzfs/zfs.git && cd zfs
sh autogen.sh
./configure \
  --disable-dependency-tracking \
  --prefix=/usr \
  --with-config=kernel \
  --sysconfdir=/etc \
  --sbindir=/usr/bin \
  --libdir=/usr/lib \
  --datadir=/usr/share \
  --includedir=/usr/include \
  --with-udevdir=/usr/lib/udev \
  --libexecdir=/usr/lib \
  --with-linux=/usr/lib/modules/6.13.6-arch1-1/build \
  --with-linux-obj=/usr/lib/modules/6.13.6-arch1-1/build
make
# save the module/*.ko files elsewhere
make distclean
./configure \
  --disable-dependency-tracking \
  --prefix=/usr \
  --with-config=kernel \
  --with-linux=/usr/lib/modules/6.13.6-arch1-1/build \
  --with-linux-obj=/usr/lib/modules/6.13.6-arch1-1/build
make
```

Maybe I'm missing something important here, but if I'm not and there's no practical difference in the end result, I'd probably stick to "if it ain't broken, don't fix it". This is my personal opinion, anyway, and if another maintainer is fine with these changes, I would not object to merging them.

And, again, the PR is helpful in any case if or when we decide to rely on the `--enable-linux-experimental` configure flag.

n0xena:
Many thanks to give my some much needed reply.
If I understand you correctly it seem to make no difference in wether the additional parameters are given or not but the result is the same. So I assume however the buildflow works the additional parameter seem to somehow get pulled in by default on Arch. In that case I'm with you and at least here it does not make sense to add them. I likely will revert them in my fork, too, so the only difference becomes to just add the experimental flag (which, at least as for now, is obsolete anyway).
From what I read zfs upstream the last full test on 6.14-rc3 passed so it's likely when 6.14 gets released later this month that 2.3.1 will work with it the same as 2.3.0 worked with 6.13 without issues. But as it was requested on my fork for now I will keep the experimental flag for someone who seem to require it for when 6.14 hits the floor and zfs takes some time again to push an update to stabilize proper 6.14 support.
As from #555 this repos target is "latest ZFS + latest Linux-LTS" anyway - which is LTS-6.12 with 6.13 as current target with 2.3.1.
As it will be tedious to keep up my branch for this PR just to drag along a blueprint for how to "inject" the experimental flag into the DKMS I rather convert this into a discussion and will close both this PR as well as #579

PR #584, opened 2025-03-10
Update to zfs version 2.3.1

l-vanni:


l-vanni:
Seems that zfs 2.3.1 tar.gz archive doesn't have a correct subfolder tree. I'll investigate and re-submit new PR

n0xena:
the possible fix was dead simple: change zfs-workdir from srcdir/zfs-pkgver to just srcdir in the src/kernel helpers
but it was reported and fixed upstream while I was still in my test build

PR #587, opened 2025-03-11
actions: don't set commit for release test from forks

### LINKED ISSUE #586: Test release pipelines don't work in PRs
Permissions issue, PRs (for good reason) from external repos can't write releases

Workaround for now: Ignore build failures of the release step from external PRs.

kerberizer:
This would fix #586 if the culprit is https://github.com/ncipollo/release-action/issues/208#issuecomment-1928581301
> I've noticed that I run into this issue in a forked repo only when I define a `commit` value.
> 
> Creating a release without a commit and the release and tags are created fine, with the commit and it fails.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,12 +32,15 @@ jobs:
         run: docker build -t archzfs-builder build-container
       - name: Run builder container
         run: docker run -e FAILOVER_RELEASE_NAME --privileged --rm -v "$(pwd):/src" archzfs-builder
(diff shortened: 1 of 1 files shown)


Doridian (approved):

kerberizer:
> Edit: Will try this in a few hours.

These must have been one of the longest "a few hours" in recorded history. It's both embarrassing and shocking, to be honest, but, anyway...

The bad news from #591 is that the fix, apparently, doesn't work, so I'm closing this.

PR #600, opened 2025-08-14
upstream: ZFS v2.3.4

kerberizer:
Tracking https://github.com/openzfs/zfs/pull/17595


n0xena:
oh wow - now I feel really dumb - I just not realized that 61_g3b64a9619 and 3b64a96 are the same and had to look up what the 61 means
I also looked into the configure script how the version string is formed and only then noticed my mistake not to check and update the META file
well - anyway - learned something new, again - I'll just bump the pkgrel when 2.3.4 drops to properly distinguish between the builds

n0xena:
guess https://github.com/openzfs/zfs/pull/17656 is the new one to track now

l-vanni:
@kerberizer I have added the sha256sum for the OpenZFS 2.3.4 release (released today)

PR #606, opened 2025-10-07
disable rc options

n0xena:
rc builds are disabled anyway - why bother to keep update unused data?
I kept them in for now so others like me can fork and enable them on thier own - or if this project decides at later point to re-enable them
it's just unnecessary updates/PRs

### DIFF
--- a/conf.sh
+++ b/conf.sh
@@ -3,11 +3,11 @@
 # FIXME: reset all kernel configs set to pkgrel=1 when this changes
 #
 openzfs_version="2.3.4"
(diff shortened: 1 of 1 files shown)


kerberizer:
FWIW, one reason I'm not moving forward on this PR is that I've been thinking it's time perhaps to consider more types of releases, since now we already know this GH-based approach works well.

One such release type could include RC-versions, which is another reason why getting rid of the versions (especially of only the versions, but not the rest of the code referring to them) might be undesirable.

That being said, let's keep it open for now if other people would like to share their opinion.

Issue #607, opened 2025-10-08
Actions build failed leading to broken repo


augustozanellato:
The Actions workflow failed tonight (https://github.com/archzfs/archzfs/actions/runs/18332429446/job/52209942396) probably due to a GitHub temporary issue, this caused `archzfs.db` to be missing from the `experimental` release and thus breaking pacman.

pacman logs:
```
$ sudo pacman -Syu
:: Synchronizing package databases...
 core is up to date
 extra is up to date
 multilib is up to date
 archzfs.db failed to download
error: failed retrieving file 'archzfs.db' from github.com : The requested URL returned error: 404
error: failed to synchronize all databases (unexpected error)
```

kerberizer:
Unfortunately, we're also hitting the problem with the recent [DoS attacks on Arch infrastructure](https://archlinux.org/news/recent-services-outages/), and more specifically on AUR, e.g., from the CI logs:
```
 > [ 5/10] RUN mkdir -p /tmp/ccm-install            &&     cd /tmp/ccm-install                  &&     wget https://aur.archlinux.org/cgit/aur.git/snapshot/clean-chroot-manager.tar.gz &&     tar -xvf clean-chroot-manager.tar.gz &&     cd clean-chroot-manager              &&     makepkg -si --noconfirm              &&     cd /tmp                              &&     rm -rfv /tmp/ccm-install:
0.144 --2025-10-08 14:10:47--  https://aur.archlinux.org/cgit/aur.git/snapshot/clean-chroot-manager.tar.gz
connected.
0.560 GnuTLS: Error in the pull function.
0.560 Unable to establish SSL connection.
```

To be honest, I'd get rid of CCM. It's a fine piece of software that makes things easier, but when I was building archzfs by myself, I dropped it from my build script. Less dependencies, less problems. This would require some work though.

kerberizer:
Fixed.

PR #611, opened 2025-10-25
ci: more resilient release workflow

### LINKED ISSUE #592: Release date doesn't change on updates
If one looks only at the releases info for our repo, they could decide we don't ever update them, because the date never changes. [experimental](https://github.com/archzfs/archzfs/releases/tag/experimental) was updated a few hours ago, but the release date is still "Oct 15, 2024". This can be particularly demoralising when browsing the main repo UI :)

Image

There doesn't appear to be a way to directly change the date of a release (see, e.g., [this](https://github.com/orgs/community/discussions/6646) somewhat old discussion).

We could delete the release and then create it anew. The latter should be possible with https://github.com/ncipollo/release-action that we currently use, maybe needing just a Markdown file with the description that we have for the release as it's a bit more complex.

However, this can be risky. What happens if the create action fails *after* we've deleted the release. Perhaps overthinking here, because the same could happen with some error producing corrupt or incomplete repo files, but still.

One way to solve this could be to create the new release under a temporary name, then delete the existing release, and finally renaming the temporary name to the final one, e.g., `experimental`.

Possibly helpful actions:
- https://github.com/marketplace/actions/edit-release
- https://github.com/marketplace/actions/delete-a-release (there's also [this](https://github.com/marketplace/actions/delete-releases), but it doesn't seem to allow explicit release name)

Caveat: I haven’t thought this through fully—there may be glaring holes I've missed.

kerberizer:
Our releases use fixed names like "experimental", which we constantly update. Because existing releases are updated "in place", there's a slight risk that we get a broken Pacman repo, e.g. #607.

This commit introduces the idea of #592: releases are created first under a temporary name before being moved to the final name. Thus, if a new release breaks, it should not affect the existing release.

This solution is probably not entirely bulletproof either, but will hopefully be an improvement over the current situation.

Tested in https://github.com/archzfs/archzfs-testing/actions/runs/18803833663

Fixes #592.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,12 @@ jobs:
           GPG_KEY_ID: "${{ vars.GPG_KEY_ID }}"
         run: docker run -e GPG_KEY_DATA -e GPG_KEY_ID -e FAILOVER_RELEASE_NAME --privileged --rm -v "$(pwd):/src" archzfs-builder
 
+      - name: Delete possible leftover temporary release
+        env:
+          GH_REPO: "${{ github.repository }}"
+          GH_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        run: gh release delete --yes "${{ env.TEMP_RELEASE_NAME }}" || true
+
       - name: Publish mainline to a temporary release
         uses: ncipollo/release-action@v1.20.0
         with:


kerberizer:
Not sure if we want to use the same workflow for the failover release. We can always add it later, of course.

Doridian (approved):
I like this. Good idea to make a release and rename. If it works (I give it a 95% chance of working)

PR #616, opened 2025-12-19
Updated ZFS to 2.4.0

Soulsuke:
Action completed successfully and the package zfs-linux allowed me to boot.

I'm not sure if I should've removed 2.4.0-rc1, but looking at older releases (2.3) it was kept, so...

### DIFF
--- a/conf.sh
+++ b/conf.sh
@@ -2,11 +2,11 @@
 #
 # FIXME: reset all kernel configs set to pkgrel=1 when this changes
 #
-openzfs_version="2.3.5"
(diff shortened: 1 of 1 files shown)


kerberizer:
Superseded by #617.

@Soulsuke thanks for the PR!

Please don't hesitate to push local branches for such changes (and let me know if for some reason you aren't able to), because the tests fail with branches from forks and we haven't been able so far to solve this (#586).

P.S. Yeah, I think it's best if we keep the RC versions until we have a firm decision what to do about such possible releases.

Issue #612, opened 2025-10-25
Trigger release on new Arch kernel packages


kerberizer:
Cannot find where somebody mentioned it, but currently, if you use the pre-built modules as opposed to `-dkms`, when new Arch kernel packages are released, you have to wait until the next scheduled rebuild at ~02:04 UTC, which isn't very convenient.

The solution shouldn't be hard:
- have a small action run frequently, perhaps every hour (just not on the hour exactly);
- the action would check if new kernel packages have been released since our last release (could be simple timestamp based check, kernel version check between our packages and kernel packages, or something else);
- if there seem to be newer packages, run the release workflow.

We could perhaps further optimize this by building and updating only the packages that need it, e.g., if a new LTS kernel is released, build just the `-lts` package. However, this would require some clever merging with the existing release.

If such workflow works reliably, we could even get rid of the scheduled nightly builds completely.

loadedice:
Thanks for the reply and feedback! I should get some time in the coming days / on the weekend to incorporate these ideas. I appreciate your fast and thorough response.

 > Indeed, we should also track linux-hardened, linux-lts, and linux-zen [...] in my opinion, important, as it may influence our design.

I agree, the current design is a bit inflexible. I'll make this change before I raise a PR.

> I'm curious if there was a specific reason to not use nvchecker

There wasn't. I just went with what I thought would be easiest to debug and get a PoC going. I'm not familiar with nvchecker, so I'll check it out, at a glance it looks good.

> We could, I guess, also track some mirror instead, possibly parsing their core.db, and then use that specific mirror in our build container. If the mirror is easily configurable, it shouldn't restrict us too much

This approach sounds good to me, I'll give it a go. This is much simpler than what I was thinking of doing, where we'd check in the build container if the version we're supposed to build for exists or not, and then bail if it doesn't 😵 

> If you're interested, I can add you as a member of the organization with the appropriate rights

Sure, sounds good! Is the idea that I would make my branch in the test repo and raise a PR from there when it's ready?

PR #626, opened 2026-02-14
Trigger archzfs release on new arch kernel release

loadedice:
This PR introduces automated triggering of ZFS package releases when new Arch Linux kernel packages are published. This works by comparing the assets (releases) in github for the repo with the latest packages in a given mirror. To avoid a sort of race condition we force the same mirror throughout the entire build process, I've already configured `vars.MIRROR_URL` to the friendly mirror mentioned in #612, so it should work as soon as it's merged.

I'm keeping this as a draft for just a bit longer, I've gone through a few iterations of this and while previous versions I've observed have worked, I want to make sure this one does too, so I need to wait for a new linux-lts, linux, linux-zen or linux-hardended release to be extra sure.

I've also added some comments to the PR to explain things / flag parts i'm not sure about.

Closes https://github.com/archzfs/archzfs/issues/612

### DIFF
--- a/build-container/Dockerfile
+++ b/build-container/Dockerfile
@@ -1,6 +1,5 @@
 FROM archlinux:base-devel
 
-# Set up the mirror from mirror-url
 ARG MIRROR_URL
 ENV MIRROR_URL=$MIRROR_URL
 


.github/workflows/release.yml
```diff
@@ -1,15 +1,23 @@
...
       - '**.sh'
       - '.github/workflows/**'
+      - '!.github/workflows/watcher.yml'
       - 'build-container/**'
       - 'packages/**'
   schedule:
     - cron: "4 2 * * *"
 
 name: Release
+run-name: ${{ inputs.kernel_version || 'Manual/Cron' }} release
```
loadedice: The run name will look like:

`linux 6.18.9.arch1-1 -> 6.18.9.arch1-2 linux-lts 6.12.70-1 -> 6.12.71-1 release` 

if there is a linux and linux-lts release.

build-container/Dockerfile
```diff
@@ -1,7 +1,16 @@
...
+ARG MIRROR_URL
+ENV MIRROR_URL=$MIRROR_URL
+
+USER root
+RUN echo "Server = ${MIRROR_URL}/archlinux/\$repo/os/\$arch" > /etc/pacman.d/mirrorlist
+
 # The following is adapted from https://github.com/archzfs/archzfs-ci/blob/master/worker/Dockerfile
 RUN pacman -Syu --noconfirm --needed python-pipx python-twisted git wget systemd-sysvcompat openresolv vi
+# Confirm the mirror is set, done after sync
+RUN pacman -Sp linux linux-lts linux-hardened linux-zen linux-headers linux-lts-headers linux-hardened-headers linux-zen-headers
```
loadedice: I only used this for debugging, and I don't actually know if the mirror here matters as much as it does in the chroot. I don't _fully_ understand the build process so I could be missing things.

build-container/Dockerfile
```diff
@@ -1,7 +1,16 @@
 FROM archlinux:base-devel
 
+# Set up the mirror from mirror-url
```
loadedice:
```suggestion
```

nvchecker.toml
```diff
@@ -0,0 +1,19 @@
+[__config__]
+oldver = "oldver.json"
+newver = "newver.json"
+
+[linux]
+source = "cmd"
+cmd = "tar -tzf core.db | sed -nE 's|^linux-([0-9].*?)/$|\\1|p'"
```
loadedice: I tried a few different approaches to get nvchecker to work here, this was the cleanest that let us force a specific mirror. I tried using the "regex" source before but it was too cumbersome, and I suspect it might not work for all mirrors.

.github/workflows/release.yml
```diff
@@ -1,15 +1,23 @@
 on:
+  workflow_dispatch:
+    inputs:
+      kernel_version:
+        required: false
+    types: [arch-kernel-update]
```
loadedice: future work could be to add an input for each artefact that is being built and only build / release that.

nvchecker.toml
```diff
@@ -0,0 +1,19 @@
+[__config__]
```
loadedice: I wasn't really sure if this belongs at the root of the repo or maybe in the github workflows dir, as it's sort of specific to that, and can't be run in isolation because it'll be missing core.db and extra.db

.github/workflows/release.yml
loadedice: Here's an example: https://github.com/loadedice/archzfs/actions/runs/22024656081

except the arrows are kinda going all over the place? I suspect it's just an issue with me setting the fallback to 0.0.0-0 and comparing to a version with a name in it (arch, or zen, or hardended) -- and when I've tested with realistic data it seems to work as expected.

loadedice:
> We're first ironing out everything in [archzfs#2](https://github.com/archzfs/archzfs-testing/pull/2), correct? I'm afraid that I'm a bit short on time these days, but I'll do my best to review it.

Ah yeah, I have confirmed it works as expected. It's been running in my fork I just neglected to merge that into the testing repo. I'll do that right now. 

IMO, there's also no rush for reviewing this PR, as it is really just a quality of life thing and its absence isn't breaking anything for any users.

> Also, once we're ready for the final review here, let's also make it a local branch, because of #586 -- otherwise we will not be able to get a successful actions run.

ahh I see, I'll close this PR then and reopen it as a local branch & it should be ready for a review.

Issue #613, opened 2025-10-28
Package files can change without changing the version number


TomiBelan:
Some time ago, I downloaded `zfs-utils-2.3.4-1-x86_64.pkg.tar.zst` from the current archzfs repo (`https://github.com/archzfs/archzfs/releases/download/experimental`). Today when I was reinstalling the package from my pacman cache, pacman complained that the file has an invalid checksum:

```
:: File /var/cache/pacman/pkg/zfs-utils-2.3.4-1-x86_64.pkg.tar.zst is corrupted (invalid or corrupted package (checksum)).
Do you want to delete it? [Y/n] n
```

When I downloaded it again, I got a different hash!

```
be5d1a909f6f5db53b2c39efaea0d766caa87fecb5d2e590a4e9c40e34c038f4  old/zfs-utils-2.3.4-1-x86_64.pkg.tar.zst
608911002c06a8d7a3aa46412211dc6572ef1a4c1e0537f51ac60006c7cce453  new/zfs-utils-2.3.4-1-x86_64.pkg.tar.zst
```

Is this intentional behavior?

I would expect that the pkg.tar.zst filename uniquely identifies a specific version, and when you rebuild it, you increase the build number (pkgrel).

(But let's also keep in mind that it should only increase if there is a meaningful change, not e.g. automatically every day.)

FYI: Comparing the two files with diffoscope shows that the only differences are in the metadata:

- timestamps in the package
- `builddate` and list of `installed = ...` package versions in `.BUILDINFO`
- `builddate` in `.PKGINFO`
- all of the above also affects `.MTREE`

TomiBelan:
Thanks for the detailed reply!

I did not use `pacman -U`. These steps should reproduce it:

1. `pacman -S zfs-utils zfs-linux`
2. `pacman -R zfs-utils zfs-linux`
3. Wait a few days
4. `pacman -Syu`
5. `pacman -S zfs-utils zfs-linux`

<details><summary>My output:</summary>

```console
$ sudo pacman -Rs zfs-linux zfs-utils
checking dependencies...

Packages (2) zfs-linux-2.3.4_6.16.10.arch1.1-1  zfs-utils-2.3.4-1

Total Removed Size:  46.33 MiB

:: Do you want to remove these packages? [Y/n] 
:: Processing package changes...
(1/2) removing zfs-linux                                        [#############################################################################################] 100%
(2/2) removing zfs-utils                                        [#############################################################################################] 100%
:: Running post-transaction hooks...
[...]

PR #632, opened 2026-02-28
Only build affected kernel package on watcher trigger

loadedice:
Followup to https://github.com/archzfs/archzfs/pull/627 (Not merged yet)

Modifies the release workflow to only build certain kernel packages based off what versions need updating. This utilises the existing "failover" method when doing a re-release, but this does feel like a bit of a hack, ideally we'd just update the files that need it, but imo this is good enough.

This means that the build is faster when only a subset of kernels needs an update and I think we might be able to take this solution to help with https://github.com/archzfs/archzfs/issues/613 if we have a good criteria on when `utils` and `dkms` _need_ to be built (I am not sure _when_ they need updates? Only when OpenZFS is bumped maybe?)

Split this into 2 different PRs because I think it's easier to review this way, smaller diffs.

This has been running in the archzfs-testing repo since https://github.com/archzfs/archzfs-testing/pull/6 and https://github.com/archzfs/archzfs-testing/pull/4 were merged (~ 1 week ago) and seems to be working well so far.

If https://github.com/archzfs/archzfs/pull/627 gets merged I will set the base branch for this to master and undraft the PR.

### DIFF
--- a/.github/workflows/watcher.yml
+++ b/.github/workflows/watcher.yml
@@ -76,6 +76,12 @@ jobs:
           if [ -n "$UPDATES" ]; then
             echo "trigger=true" >> $GITHUB_OUTPUT
             echo "updates=$UPDATES" >> $GITHUB_OUTPUT
+
+            # Parse each kernel explicitly
+            echo "$UPDATES" | grep -qE '(^| )linux ' && echo "build_linux=true" >> $GITHUB_OUTPUT
+            echo "$UPDATES" | grep -q 'linux-lts' && echo "build_linux_lts=true" >> $GITHUB_OUTPUT
+            echo "$UPDATES" | grep -q 'linux-hardened' && echo "build_linux_hardened=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


.github/workflows/watcher.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           cat oldver.json
 
       - name: Check for updates
+        continue-on-error: true # Continue if grep doesn't match all.
```
loadedice, line 68: it's either this, or add `|| true` to all the greps below because if any command exits with non 0 status it fails the whole pipeline

build-container/entrypoint.sh
```diff
@@ -60,13 +60,18 @@ build() {
...
+
     if [ -z "${FAILOVER_REPO_DIR}" ]; then
-        echo "No failover repo available, failing because of: $1"
+        echo "No failover repo available, failing because of: $failed_pkg"
         exit 1
     fi
 
-    failed_pkg="$1"
-
     set +x # This gets way to verbose
```
loadedice, line 75: just a little typo i noticed after making this PR
```suggestion
    set +x # This gets way too verbose
```

PR #642, opened 2026-06-13
update to zfs 2.4.3

n0xena:
ZTS result: https://github.com/n0xena/archzfs/pull/38#issuecomment-4699146036
note: 2.4.3 doesn't support linux 7.1 yet - so 2.4.4 with support for linux 7.1 is to be expect rather quick


kerberizer:
Superseded by #649, though it's blocked anyway.

Thank you very much for the contribution!

PR #649, opened 2026-07-12
Update OpenZFS to 2.4.3

kerberizer:
## Summary

- update OpenZFS from 2.4.2 to 2.4.3 using the upstream release checksum
- remove `zfs-utils` handling for `/etc/sudoers.d/zfs`, which upstream removed
- enable the Linux 7.1 support already present in OpenZFS 2.4.2 and 2.4.3
- reset the stable utilities `pkgrel` for the new upstream version
- clarify that RC and Git package inputs are not part of the current production build

## Rationale

OpenZFS 2.4.3 removes `/etc/sudoers.d/zfs`. The existing `backup` entry would consequently refer to a file absent from the package, while the permission adjustments would target paths that are no longer installed. This packaging adjustment must accompany the version update because OpenZFS 2.4.2 still provides that file.

The stable `zfs-utils` `pkgrel` is reset from 2 to 1, as required for a new upstream version.

OpenZFS has confirmed that the necessary Linux 7.1 compatibility changes are present starting with 2.4.2. The 2.4.2 release archive accidentally carries a `Linux-Maximum` value of `99.99`, while 2.4.3 restores the pre-validation limit of 7.0. Upstream subsequently validated the final Linux 7.1 release with ZTS and merged a metadata-only support update.

The generated prebuilt and DKMS recipes therefore normalize `Linux-Maximum` to 7.1 only for OpenZFS 2.4.2 and 2.4.3. The expected source values are checked, and RC, Git, and future releases remain unaffected. A reminder beside the configured OpenZFS version calls for removing the workaround once the selected release carries the required support marker itself.

This supersedes #641 and #642. It combines their useful changes atomically, adds the missing `pkgrel` reset, and retains attribution to their authors.

## Upstream verification

- OpenZFS 2.4.3 release: https://github.com/openzfs/zfs/releases/tag/zfs-2.4.3
- Signed checksums: https://github.com/openzfs/zfs/releases/download/zfs-2.4.3/zfs-2.4.3.sha256.asc
- Removal of `/etc/sudoers.d/zfs`: https://github.com/openzfs/zfs/pull/18626
- Linux 7.1 compatibility confirmation: https://github.com/openzfs/zfs/issues/18760#issuecomment-4919127088
- Linux 7.1 metadata update and ZTS result: https://github.com/openzfs/zfs/pull/18682
- Fix for the accidental `99.99` release metadata: https://github.com/openzfs/zfs/pull/18531

The configured SHA-256 is:

`1f08f2d154f5189b5f1382848a32667b3d34066145b474c49cd3d41a5fba59a7`

## Validation

Local checks:

- `bash -n` on all changed shell templates
- generated prebuilt and DKMS PKGBUILDs pass `bash -n`
- the gated metadata normalization produces 7.1 from the official 2.4.2 and 2.4.3 release archives
- `git diff --check`

The same-repository [Test run](https://github.com/archzfs/archzfs/actions/runs/29250082786) completed successfully and published the mutable `testing` release. It built OpenZFS 2.4.3 packages for utilities, DKMS, Linux 6.18 LTS, and the current Linux 7.1 standard, hardened, and zen kernels without failover. The resulting kernel packages all require `zfs-utils=2.4.3`.

The published `zfs-dkms` package contains `Linux-Maximum: 7.1`. Its source was also extracted and successfully configured and compiled into `spl.ko` and `zfs.ko` against Linux 7.1.3 headers without an experimental-kernel flag.

ArchZFS validation here covers package generation and compilation; it does not load the modules or exercise filesystem operations. Runtime confidence for Linux 7.1 relies on the upstream ZTS result linked above.

No generated package submodules were modified. RC and Git packages remain unpublished and are not enabled by this change.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -317,7 +317,8 @@ for (( i = 0; i < ${#modes[@]}; i++ )); do
     for func in "${update_funcs[@]}"; do
         debug "Evaluating '${func}'"
 
-        # skip git packages if -s was used
+        # Production does not publish RC or Git packages. Review their
+        # generators and source-version assumptions before enabling them.
         if [[ ${only_stable} -eq 1 && ( ${func} =~ git_pkgbuilds$ || ${func} =~ rc_pkgbuilds$ ) ]]; then
             debug "Skipping '${func}' (non stable)"
             continue
--- a/conf.sh
+++ b/conf.sh
@@ -2,11 +2,14 @@
 #
 # FIXME: reset all kernel configs set to pkgrel=1 when this changes
 #
-openzfs_version="2.4.2"
+openzfs_version="2.4.3"
+
+# RC packages are currently disabled. These retained values are not production
+# release inputs; review the complete RC path before updating or re-enabling it.
 openzfs_rc_version="2.4.0-rc1"
 
 # The OpenZFS source hashes are from github.com/openzfs/zfs/releases
-zfs_src_hash="7e260d0e6af295bea4c5e241cac0a1aef07b58d8dd8035f7898ade3b1bbec78f"
+zfs_src_hash="1f08f2d154f5189b5f1382848a32667b3d34066145b474c49cd3d41a5fba59a7"
 zfs_rc_src_hash="0068102a4162d7445b80218b882ff54e1acf3cfbeef909d53bd984ddcd9339b1"
 
 zfs_initcpio_install_hash="d19476c6a599ebe3415680b908412c8f19315246637b3a61e811e2e0961aea78"
--- a/conf.sh
+++ b/conf.sh
@@ -1,6 +1,8 @@
 # OpenZFS stable version
 #
 # FIXME: reset all kernel configs set to pkgrel=1 when this changes
+# FIXME: remove the Linux-Maximum workaround in src/zfs/PKGBUILD.sh and
+# src/zfs-dkms/PKGBUILD.sh once the selected release declares Linux 7.1 or later
 #
 openzfs_version="2.4.3"
 
--- a/src/kernels/_utils.sh
+++ b/src/kernels/_utils.sh
@@ -3,7 +3,7 @@ mode_name="utils"
 mode_desc="Select and use the utils packages"
 
 # version
-pkgrel="2"
+pkgrel="1"
 
 # Version for GIT packages
 pkgrel_git="1"
--- a/src/zfs-dkms/PKGBUILD.sh
+++ b/src/zfs-dkms/PKGBUILD.sh
@@ -19,6 +19,13 @@ groups=("${archzfs_package_group}")
 conflicts=("zfs" "zfs-headers" "spl" "spl-headers")
 ${zfs_replaces}
 
+prepare() {
+    cd "${zfs_workdir}"
+    # openzfs/zfs#18682 confirms Linux 7.1 support after the 2.4.3 release.
+    sed -i 's/^Linux-Maximum: 7.0$/Linux-Maximum: 7.1/' META
+    grep -qx 'Linux-Maximum: 7.1' META
+}
+
 build() {
     cd "${zfs_workdir}"
     ./autogen.sh
--- a/src/zfs-dkms/PKGBUILD.sh
+++ b/src/zfs-dkms/PKGBUILD.sh
@@ -21,9 +21,14 @@ ${zfs_replaces}
 
 prepare() {
     cd "${zfs_workdir}"
-    # openzfs/zfs#18682 confirms Linux 7.1 support after the 2.4.3 release.
-    sed -i 's/^Linux-Maximum: 7.0$/Linux-Maximum: 7.1/' META
-    grep -qx 'Linux-Maximum: 7.1' META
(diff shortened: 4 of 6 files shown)


kerberizer:
We'll indeed have to wait for OpenZFS 2.4.4 for this, but it was a good test. I'll keep this open, but downgraded to draft, for record-keeping and reference for the small additional changes.

kerberizer:
Correction to my note above: we do not need to wait for 2.4.4 after all. The 2.4.2 release archive was accidentally produced with a `Linux-Maximum` value of `99.99`, which explains why our existing DKMS package already built on 7.1. More importantly, upstream has confirmed that the required compatibility changes are present since 2.4.2, and its later 7.1 enablement was only a metadata update after testing the final kernel.

The workaround is now limited to 2.4.2 and 2.4.3 rather than enabling arbitrary experimental kernels. The final test run built every current package family as 2.4.3 without failover, and the published DKMS source also compiled successfully against 7.1.3 headers. I have updated the PR description with the upstream references and validation details.

kerberizer:
For the record, this PR was the first concrete case that gave me real confidence that AI assistance could help me regain a sustainable maintainership pace.

It began with two useful human contributions, #641 and #642. They identified real changes that were needed, and their authors retain attribution here. But the two changes had to be combined atomically, reviewed against the current build and release process, and supplemented with several details.

One of those details was resetting `pkgrel` to `1` for the new OpenZFS release. The slightly embarrassing part is that this was not a new oversight.

When we updated to OpenZFS 2.4.1 in February, I accepted the change without noticing that `pkgrel` had remained at `2`. I then pulled the same commit into my own #631, specifically so that our GitHub Actions workflow could test it, and still did not notice. The following 2.4.2 update inherited the same value. Thus, after the legitimate `2.4.0-1` and `2.4.0-2`, we published `2.4.1-2` and `2.4.2-2`, although both new upstream versions should have started at `-1`.

To be completely clear, I consider this my responsibility. I reviewed and merged the change, then handled it again in my own PR, and failed to spot the problem both times.

Fortunately, this particular mistake was mostly harmless. But it is also a good example of why I am reluctant to review PRs unless I can devote sufficient time, energy, and concentration to them. The next small detail that gets overlooked may not be harmless.

The work on this PR provided another useful example. The first CI run appeared to complete successfully, but the assistant noticed that the relevant builds had actually fallen back to our backup packages because of the apparent version incompatibility. We stopped at that point, returned the PR to draft, and initially concluded that we would have to wait for OpenZFS 2.4.4.

The following day, I noticed that our existing DKMS package was nevertheless building successfully against Linux 7.1. That did not fit the initial conclusion, so the assistant and I investigated further. We established from the upstream changes and release archives that the actual compatibility code was already present, while the remaining obstacle was release metadata. With some guidance from me about the packaging approach and acceptable scope of the workaround, we arrived at the current solution.

The workaround is deliberately limited to OpenZFS 2.4.2 and 2.4.3, checks the expected upstream metadata, and does not enable arbitrary experimental-kernel builds. The final test built all current package families as OpenZFS 2.4.3 without falling back to the backup packages. The generated DKMS source was also compiled against Linux 7.1.3 headers.

Could another human developer have done all of this without AI assistance? Absolutely. Nothing here requires some supposedly magical machine ability.

The practical questions are whether somebody actually had the time and concentration to do it, and whether the work would otherwise have been completed within a reasonable period. The earlier PRs had remained open, nobody else had undertaken the combined integration and investigation, and I would have found it very difficult to devote enough uninterrupted time to all of it unaided.

The assistant handled much of the routine but concentration- and time-intensive work: comparing changes, following references, inspecting generated results, and iterating over possible solutions. I reviewed the evidence, supplied the project-specific context and judgment, decided when we had to stop, and remain responsible for the proposed result.

I do not think this proves that AI is more dependable than humans. The assistant required supervision and correction elsewhere. Conversely, I had personally overlooked the `pkgrel` mistake across two interactions with the same change, while the assistant noticed it during this work. Humans and AI systems have different failure modes, and neither should be trusted without review.

What this PR did demonstrate to me is that carefully supervised assistance can make work realistically achievable that might otherwise remain unfinished. It does not remove the need for human review -- as the requested reviews on this PR should make sufficiently clear -- but it gives me some confidence that I can regain a reasonable maintenance tempo without lowering the standard of care I believe this project requires.

l-vanni (approved):

n0xena:
Thank you for this insightful response to this.
Also thank you very much for still mention me as contributor although my "contributions" aren't much more than merely have won the "race" to "Who opens the PR first after upstream release?" once or twice.
I know it's not much but that's my way of giving back to Linux, Arch, ZFS to keep it "free" (as in: free of charge for others) - simply because I learned: I doesn't matter how much but it's people like us with projects like this who keep this whole momentum of Linux alive. And looking at how much Valve pushed gaming on Linux by investing heavily into wine - and also how M$ keeps screwing up - maybe one day will be "The day of Linux on desktop".

Keep it up everybody - and if it takes time and maybe some 3rd party assistence - so be it. I guess from #555 the focus of this project was made clear by lengthy back and forth.

PR #648, opened 2026-07-12
Separate release-body publication and clarify signing guidance

kerberizer:
## Motivation

Manual package signing is a transitional trust procedure, not the intended long-term user setup. Release-body changes should be published independently of package publication, rather than waiting for a later package release to update the notes or entering the production signing environment.

The package-release cancellation hazard discovered during review was fixed separately in #652 (`bed751e`, merged as `1841639`). Its [production run](https://github.com/archzfs/archzfs/actions/runs/31533159783) completed successfully.

## Changes

- Add a dedicated metadata workflow for the `experimental` release body.
- Restrict its release mutation to `gh release edit experimental --notes-file`.
- Check out the body explicitly from current `master`.
- Use separate `release-body` concurrency with `cancel-in-progress: false`.
- Grant only `actions: read` for package-run inspection and `contents: write` for release editing.
- Do not use the `Release` environment or signing material.
- Reconcile the body after every package `Release` completion, regardless of conclusion.
- Narrow the package workflow's workflow-file path filter from `.github/workflows/**` to `.github/workflows/release.yml`.
- Describe manual key import and local signing as optional and transitional.
- Display the imported key's fingerprint before local signing.
- Retain `SigLevel = Never` only as a temporary compatibility option without authenticity verification.

## Release behavior

Package publication continues to use the body from its own checkout. If that checkout is older, the completion-triggered metadata workflow reapplies the body from current `master`.

If a package release is active, metadata publication defers instead of racing the delete-and-rename promotion. A failed metadata edit with no active package run remains a real failure, including when the public `experimental` release is unexpectedly missing.

Notes-only updates do not replace assets, move tags, alter release state, or refresh GitHub's displayed package-release date.

Merging this PR will trigger one package release because `release.yml` itself changes. Subsequent body-only changes will publish through the dedicated metadata workflow, while unrelated workflow-file changes will no longer trigger package publication.

## Current-master update

Candidate `dba48b348ec540d409534127a09861a9c6b72c58` merges the previous PR head `7ceed71e9be6f09502c3359cd1111cd91ffc9dc7` into production baseline `926efc82511da72774c02d8e9af338a64c9093f1`.

Production `master` subsequently advanced to `379a8cd06e36dbb8855d8158c582c6eca5cdcc78` through #665. GitHub reports the PR as conflict-free, and `git merge-tree` against that commit is clean, so no further branch update is needed.

[...]

PR #652, opened 2026-08-11
ci: prevent cancellation during release promotion

kerberiel:
## Motivation

While reviewing #648, I found that the package `Release` workflow can cancel an active run when a newer qualifying push or scheduled run arrives.

#611 introduced create-then-promote publication so an incomplete upload does not replace the working repository and each package generation receives a meaningful publication date. The resulting promotion sequence uploads `_experimental`, deletes the existing `experimental` release, and then renames the temporary release. If automatic cancellation occurs between deletion and rename, the public Pacman repository can remain unavailable until manual recovery or a later release run reaches promotion.

The same policy also protects the in-place `failover` replacement, where interruption during `removeArtifacts: true` can leave the fallback release empty or partial.

## Change

Set `cancel-in-progress: false` for the package `Release` workflow and document the corresponding operational invariant.

GitHub still permits only one running and one pending member of the concurrency group by default. A newer run replaces an older pending run, so this does not create an unbounded queue.

No package inputs, signing behavior, release names, assets, tags, triggers, promotion order, or failover behavior change.

## Trade-off and Follow-up

This is a containment measure rather than the intended final design. A superseded build may now continue for roughly 20-25 minutes instead of being canceled.

A future redesign should separate cancellable build work from a serialized, non-cancellable signing and publication stage. That split also needs to pin failover input to a complete generation: the builder currently reads the mutable failover database early and may download fallback packages much later, while publication replaces failover assets in place. The independently concurrent `Test` workflow has the same pre-existing mutable-failover race.

#653 tracks that broader redesign, including bounded build execution, safe promotion, and immutable failover input. Until those boundaries are separated safely, protecting the entire workflow is the smallest reliable fix while retaining the properties introduced by #611.

## Operational Impact

Merging this PR changes `release.yml`, so it will trigger one production package release. The merged workflow definition will prevent that run from automatically canceling an older member of the same concurrency group.

The delete-and-rename sequence still has a brief non-atomic interval. Manual cancellation, runner loss, API failures, and the existing six-hour job timeout remain outside the protection provided here.

## Validation

- `actionlint .github/workflows/release.yml`
- `git diff --check`
- No privileged local package build, because package and builder inputs are unchanged
- No deliberate cancellation or overlap test against production

I also reviewed the README, architecture and roadmap documentation, release body, organization profile, and Wiki. They require no update for this narrowly scoped concurrency change.

## Review Focus

Please verify the concurrency behavior, the delete-to-rename failure window, and the decision to keep the broader cancellable-build redesign separate.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ name: Release
 
 concurrency:
   group: release
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 permissions:
   contents: write
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -93,6 +93,11 @@
 - Preserve create-then-promote publication for fixed-name release channels. In
   addition to reducing exposure to incomplete uploads, creating a fresh release
   updates the date shown by GitHub; updating assets in place does not.
+- While fixed-name package publication uses delete-then-rename promotion, its
+  release-mutating critical section must not be canceled. The current workflow
+  protects the whole run as a containment measure; make build work cancellable
+  only after separating it from publication and pinning failover inputs to a
+  complete generation.
 - Preserve failover database/package signature verification when changing the
   builder. A downloaded package is not eligible for reuse merely because its
   filename matches.


kerberizer (approved):
**Verdict: correct and well-scoped. Approve, ideally with one follow-up (job timeout) added here or tracked.**

## What I verified

- `concurrency.group: release` is a constant, so all `push`-to-`master` and `schedule` runs share one group — flipping `cancel-in-progress` to `false` genuinely serializes the publication critical section (`.github/workflows/release.yml:14-16`).
- The stated queueing semantics are accurate: with `cancel-in-progress: false` GitHub allows one running + one pending member and cancels the *previously* pending run when a newer one arrives. No unbounded queue, and the pending run always carries the newest commit.
- The failure window is real. `Delete the existing mainline release` → `Rename the temporary release to mainline` (`release.yml:56-70`) leaves `experimental` nonexistent in between; the Pacman repo 404s until a later run promotes.
- `test.yml` is in a separate group (`test`) and writes only `testing`, so no cross-workflow write conflict is introduced.
- Validation reproduces: `actionlint .github/workflows/release.yml` clean, YAML parses, `git diff --check master...HEAD` clean. PR checks green (`Test release` passed, 23m49s — consistent with the 20–25 min figure in the description).
- The AGENTS.md rationale matches the code: `build-container/entrypoint.sh:24` fetches the failover DB early, `:85` downloads packages much later, and the `Release failover` step replaces assets in place (`allowUpdates` + `removeArtifacts`). The "pin failover inputs to a complete generation" precondition is well founded.

## Findings

**1. No `timeout-minutes` on the release job — the new blocking gap.** There is no `timeout-minutes` anywhere in `.github/workflows/`, so the job inherits the 6-hour default. Previously, a hung run was implicitly unwedged: the next push or nightly schedule canceled it. Now a wedged run holds the group for up to 6 hours, and since only one pending run is retained, intervening scheduled runs are dropped. `release.yml` also has no `workflow_dispatch`, so recovery requires a manual cancel in the UI. Adding `timeout-minutes: 60` to the `release` job restores a bounded worst case without weakening the containment (a timeout kill inside the critical section is no worse than today's cancel — and considerably rarer).

**2. The description undersells its own coverage.** The motivation cites only the `experimental` delete→rename window, but the change also protects the `Release failover` step, which deletes every failover asset before re-uploading (`removeArtifacts: true`). A cancel there previously could have left `failover` empty — arguably a worse outcome, since it is the fallback path. Worth one sentence in the PR body.

**3. Pre-existing, correctly left out of scope:** a `Test` run reading failover packages (`entrypoint.sh:85`) can race a `Release` run rewriting those same assets — the groups are independent. Not introduced here, and the AGENTS.md note already points at the right fix. Fine to leave for the redesign, but naming it in the follow-up section would sharpen the scope of that work.

## Minor

- The `Operational Impact` claim that merging triggers one production release is correct — the `.github/workflows/**` path filter matches; `AGENTS.md` alone would not have.
- AGENTS.md wording and placement are consistent with the surrounding `Operational Safety` bullets; line lengths conform.

No blocking issues. Finding 1 is the only one I'd want resolved before or shortly after merge.

kerberiel:
Thanks for the thorough review. I agree that bounding a wedged build matters, but I am leaving the proposed job-wide 60-minute timeout out of this containment PR: while build and publication share one job, that timeout could itself terminate the delete-then-rename critical section.

I opened #653 to track the coupled redesign. Its acceptance criteria include a bounded cancellable build stage, serialized non-cancellable publication, pinned failover generations, and safer promotion. I also updated this PR description to cover the in-place failover replacement and the pre-existing `Test`/`Release` failover race.

kerberiel:
Post-merge verification completed successfully.

- Merge commit: `18416393a21581ad96ca3796fac9cd84ae7f6e8f`
- Production Release: https://github.com/archzfs/archzfs/actions/runs/31533159783
- The run completed all build, promotion, failover, and tagging steps successfully in 20m04s.
- `experimental` and `failover` each contain the same complete 30-asset generation by name, size, and digest.
- Both annotated fixed tags resolve to the merge commit.
- No `_experimental` release or tag remains.
- The public `archzfs.db` and `archzfs.db.sig` URLs return HTTP 200.

The broader cancellable-build and safe-publication redesign is tracked in #653.

PR #656, opened 2026-08-22
Update to zfs 2.4.4

n0xena:
update to zfs 2.4.4
ZTS finished successful without any non-expected skips/failures: https://github.com/n0xena/archzfs/pull/42#issuecomment-5382271397 on 7.1.8-3

PR #657, opened 2026-08-23
Update conf.sh to zfs 2.4.4

l-vanni:
Supersedes #656. Closes #656.

### DIFF
--- a/conf.sh
+++ b/conf.sh
@@ -1,8 +1,6 @@
 # OpenZFS stable version
 #
 # FIXME: reset all kernel configs set to pkgrel=1 when this changes
-# FIXME: remove the Linux-Maximum workaround in src/zfs/PKGBUILD.sh and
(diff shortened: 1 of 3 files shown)


l-vanni:
I recreated the same change from an internal archzfs/archzfs branch so the required CI checks can run with the necessary permissions.

kerberiel (approved):
Verified the 2.4.4 source hash against upstream and rechecked the follow-up against #649: the Linux 7.1 metadata workaround was intentionally scoped to 2.4.2 and 2.4.3, while 2.4.4 declares support through Linux 7.2, so its removal is appropriate. CI passed; utils, DKMS, standard, hardened, and zen built as 2.4.4. LTS used signed failover only because linux-lts=6.18.45-2 was no longer available from the Arch repositories, before source preparation, so that failure does not implicate this change. I found no remaining issues.

kerberizer (approved):
LGTM

Issue #655, opened 2026-08-18
Is libunwind a required dependancy?


reaperx7:
As the person currently maintaining the Install ArchLinux on ZFS page of the ArchWiki, the fact that zfs-utils seems to require libunwind has been called into question since the zfs-utils package from the AUR has a build flag for --without-libunwind. Is there a specific reasoning for usage of libunwind by zfs-utils that we need to be aware of?

kerberiel:
Sorry for the delayed response. I wanted to verify the published artifact and have a concrete, tested packaging change ready before answering.

Short answer: no. The signed ArchZFS `zfs-utils 2.4.4-1` package currently published does not require or link libunwind.

I checked the package itself, not only the recipe. Its SHA-256 is `c8a32012d041dd776d9e4a59ab1b121c3d8dce6fe1b77fe569ec56bd02e0d857`, and its signature verifies under the ArchZFS release-key fingerprint `3A9917BF0DED5C13F69AC68FABEC0A1208037BE9`. None of its 92 ELF objects has a `libunwind` `DT_NEEDED` entry, and `.BUILDINFO` confirms that libunwind was absent from the clean build environment. The [producing run](https://github.com/archzfs/archzfs/actions/runs/33601604827) records `checking for libunwind... no` and `checking for library containing backtrace... none required`.

OpenZFS uses libunwind only to provide richer userspace crash diagnostics. Without it, OpenZFS retains its glibc `backtrace()` fallback, which the published libraries use. This remains relevant to `VERIFY*` failure paths, but normal `zfs` and `zpool` operation does not depend on libunwind. The three backtrace implementations are visible in the [pinned OpenZFS source](https://github.com/openzfs/zfs/blob/zfs-2.4.4/lib/libspl/backtrace.c).

There is one important qualification: the current production recipe leaves upstream auto-detection enabled. The present package selected the glibc implementation because its clean build environment did not contain libunwind, rather than because the recipe explicitly selected that policy.

[PR #660](https://github.com/archzfs/archzfs/pull/660) proposes to make the existing outcome deterministic with `--without-libunwind`.

The investigation also exposed a separate packaging defect: the published `zfs-utils` package declares no runtime dependencies. The same PR adds direct runtime dependencies and moves the approximately 35 MiB ZFS Test Suite payload into an optional `zfs-tests` child intended for disposable test systems.

[...]

PR #658, opened 2026-08-24
Add opt-in TPM2 key provider for encrypted ZFS root

radupotop:
# Summary

This PR adds an opt-in initramfs key-provider path for ArchZFS'
mkinitcpio integration, based on the alternate key-loading contract already
documented by OpenZFS.

It also ships a TPM2 provider example using `systemd-creds`, so a native ZFS
encrypted root dataset can be unlocked unattended during early boot.

The normal ZFS hook remains generic. Users who do not configure a provider get
no TPM behavior, no TPM payload in their initramfs, and no new required TPM
dependency.

# Motivation

The target use case is a native-encrypted ZFS root filesystem on a machine that
needs unattended reboot.

For example, a small system with only two M.2 slots can use both drives for a
mirrored ZFS pool. In that layout there is no third drive for a separate OS
disk, so the operating system lives on the same redundant pool. That is a good
fit for ZFS snapshots, send/receive, and root filesystem recovery, but the root
dataset still needs to be unlocked before `/` can be mounted.

This change keeps that native ZFS encryption layout and adds a TPM-assisted
boot path without requiring LUKS-under-ZFS, a separate key disk, or local
site-specific boot scripts.

# Design

The mkinitcpio runtime hook now looks for provider scripts at:

```text
/etc/zfs/initramfs-tools-load-key
/etc/zfs/initramfs-tools-load-key.d/*
```

If present in the initramfs, providers run before the normal `zfs load-key`
fallback. They receive:

- `DATASET`
- `ENCRYPTIONROOT`
- `KEYLOCATION`
- `BOOT_DATASET`
- `ZFS`
- `ZPOOL`

Providers are sourced in a subshell. That keeps provider `return`, `exit`,
variable changes, and runtime failures isolated from the BusyBox `/init`
process. Provider failure is non-fatal and the existing passphrase fallback
remains available.

The hook passes its already resolved boot dataset to providers. This avoids a
second provider-side parser for `/proc/cmdline`, and keeps explicit bootloader
overrides such as `root=ZFS=tank/root` aligned with the dataset the hook will
actually mount.

# TPM2 Provider

The TPM2 provider example is installed as documentation at:

```text
/usr/share/doc/zfs-utils/initramfs-tools-load-key.d/tpm2-systemd-creds
```

It is not enabled by default. To opt in, copy it into `/etc/zfs`:

```sh
install -Dm644 \
    /usr/share/doc/zfs-utils/initramfs-tools-load-key.d/tpm2-systemd-creds \
    /etc/zfs/initramfs-tools-load-key.d/tpm2-systemd-creds
```

The provider decrypts `/etc/zfs/zfs-key.creds` with `systemd-creds` and pipes
the result to:

```sh
zfs load-key -L prompt "${ENCRYPTIONROOT}"
```

Using `-L prompt` ensures the piped credential is consumed even when the
dataset's configured `keylocation` is not `prompt`.

The TPM2 provider only attempts to unlock the encryption root for the boot root
dataset. Other encrypted datasets continue through the normal key-loading path,
although datasets that inherit the same encryption root may also become
available once that root key is loaded.

The provider is not a mkinitcpio hook. Users keep `zfs` in `HOOKS`; they do not
add a separate TPM2 hook or mkinitcpio's `systemd` hook for this provider.

# Initramfs Payloads

Provider scripts under `/etc/zfs` are copied into the initramfs only when they
exist.

For the TPM2 provider specifically, the install hook conditionally adds:

- `systemd-creds`
- TPM kernel modules under `/drivers/char/tpm/`
- required `libtss2-*` libraries discovered under `/usr/lib`

The old ZFS-specific provider payload manifest was removed. Users should use
mkinitcpio's existing `FILES=()` mechanism for provider payloads, for example:

```bash
FILES+=(/etc/zfs/zfs-key.creds)
```

# Packaging

This branch was updated after the post-#660 `zfs-utils` split package work.

The current package layout is:

- stable `zfs-utils` is bumped to `2.4.4-3`
- the TPM2 provider source is included in the package sources
- the provider checksum is tracked in `conf.sh`
- `build.sh -U` refreshes the provider checksum
- the provider example is installed only by `package_zfs-utils()`
- `tpm2-tss` is an optdepend only for the `zfs-utils` child package
- no TPM provider file, install step, or optdepend is added to `zfs-tests`
- `systemd` is not a TPM-specific optdepend because it is already a direct
  `zfs-utils` dependency after #660

The same source template also generates the `zfs-utils-git` package recipe:

- `pkgbase="zfs-utils-git"`
- child packages are `zfs-utils-git` and `zfs-tests-git`
- the provider is installed only by `package_zfs-utils-git()`
- `tpm2-tss` is scoped only to `zfs-utils-git`

# Documentation

The `mkinitcpio -H zfs` help text now documents:

- the provider loading contract
- how to enable the packaged TPM2 provider example
- the `tpm2-tss` requirement
- use of the existing BusyBox initramfs
- that mkinitcpio's `systemd` hook is not required
- credential creation with `systemd-creds`
- use of `FILES=()` for `/etc/zfs/zfs-key.creds`
- optional `zfs_tpm2_credential=` and `zfs_tpm2_credential_name=` overrides
- root-dataset-only TPM2 unlock behavior
- inherited encryption root behavior
- keeping the original passphrase available
- recreating credentials after `zfs change-key`, TPM clear, or TPM SRK changes

# Validation

Local checks:

```sh
dash -n src/zfs-utils/zfs-utils.initcpio.hook \
    src/zfs-utils/zfs-utils.initcpio.tpm2-provider

bash -n build.sh conf.sh src/kernels/_utils.sh \
    src/zfs-utils/PKGBUILD.sh \
    src/zfs-utils/zfs-utils.initcpio.install

git diff --check master
```

The stable and `-git` package recipes were regenerated in an Arch Linux Podman
container and inspected. The generation flow used `podman run -it archlinux`
with `--userns=keep-id:uid=0,gid=0`, so the repo scripts ran as root inside
the Arch container while generated files on the host remained owned by the
normal user.

Generated package directories inspected:

- `packages/_utils/zfs-utils`
- `packages/_utils/zfs-utils-git`

Generated hook, install, and provider files matched the source templates
byte-for-byte. Generated package diffs passed `git diff --check`.

The stable split packages were built locally with:

```sh
env PKGDEST=/tmp/ makepkg -fcC --nosign
```

Resulting archives:

- `/tmp/zfs-utils-2.4.4-3-x86_64.pkg.tar.zst`
- `/tmp/zfs-tests-2.4.4-3-x86_64.pkg.tar.zst`

Archive inspection confirmed:

- `zfs-utils` contains the ZFS mkinitcpio hook
- `zfs-utils` contains the ZFS mkinitcpio install hook
- `zfs-utils` contains the TPM2 provider documentation/example
- packaged hook, install hook, and provider match the generated files
- `tpm2-tss` appears as a `zfs-utils` optdepend
- `zfs-tests` contains no TPM provider material
- the removed custom manifest is not packaged
- the credential payload is not packaged

Target-machine validation was also performed on the real TPM/ZFS system.

A screenshot from the booted machine shows:

- `zpool status -vPL`
- `zfs list -t all`
- `lsinitcpio -lv /boot/initramfs-linux-lts.img`
- `zfs get encryption,keyformat,keylocation,mountpoint`

Three validation videos were recorded:

1. unattended unlock with TPM2
2. missing `zfs-key.creds` falling back successfully to passphrase entry
3. missing TPM2 provider script falling back successfully to passphrase entry


kerberiel (changes requested):
The use case is valid: unattended unlocking of a native-encrypted ZFS root is useful for headless and remote systems, and ArchZFS already owns downstream mkinitcpio integration. I am not requesting changes because I think TPM unlocking is inherently out of scope.

The current implementation cannot merge, however. It breaks existing package and initramfs generation, imposes TPM-specific costs on all users, and only works correctly for a narrower configuration than the PR currently implies.

**Blocking issues**

1. **The changed initcpio files have stale checksums.**

   [`PKGBUILD.sh` emits the hook and install-file checksums from `conf.sh`](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/PKGBUILD.sh#L18-L21), but both files change without updating [those values](https://github.com/archzfs/archzfs/blob/2b52ad7bdc3375cf9104813f5f6cbbc9c26298b5/conf.sh#L15-L17). The release path uses `build.sh -u`, not the checksum-refreshing `-U`, so the mandatory `zfs-utils` build will fail rather than repair this automatically.

2. **`pkgrel` is not bumped.**

   [`pkgrel` remains `1`](https://github.com/archzfs/archzfs/blob/2b52ad7bdc3375cf9104813f5f6cbbc9c26298b5/src/kernels/_utils.sh#L5-L9) despite materially changed package metadata and installed files. Republishing different `2.4.4-1` packages would reuse an existing package identity and give installed users no upgrade path.

3. **Ordinary ZFS-root users get a new mkinitcpio failure.**

   The install hook [unconditionally calls `add_file` for `/etc/zfs/zfs-key.creds`](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/zfs-utils.initcpio.install#L30-L41). Current mkinitcpio records a builder error when that file does not exist and exits non-zero. Users who have the `zfs` hook configured but have not opted into TPM unlocking must continue to get a successful build with no TPM payload.

4. **TPM support becomes mandatory for every package user.**

   The PR adds [`tpm2-tss` as a global dependency](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/PKGBUILD.sh#L9-L12) and adds TPM/systemd binaries and libraries to every configured ZFS initramfs. ZFS itself does not require these. [`systemd-analyze` is also included](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/zfs-utils.initcpio.install#L12-L28) despite being described as debug-only.

   At minimum, a direct TPM implementation needs explicit opt-in, optional package metadata, and conditional image contents. Hard-coded TSS sonames should also be replaced with mkinitcpio's normal dynamic dependency resolution.

5. **The pipeline ignores its input for URI `keylocation` values.**

   The hook [pipes plaintext to `zfs load-key` without `-L`](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/zfs-utils.initcpio.hook#L85-L96). In OpenZFS 2.4.4, piped stdin is consumed when the configured key location is `prompt`; for `file://`, `https://`, and other URI locations, `zfs load-key` invokes the URI handler instead and never reads the pipe. A piped provider must use `zfs load-key -L prompt` so its behavior is independent of the dataset's configured location.

6. **One credential is applied to every encryption root.**

   The fixed [`/etc/zfs/zfs-key.creds`](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/zfs-utils.initcpio.hook#L80) and credential name are tried for every encryption root and again on every interactive retry. That covers a single-root installation, but not systems with distinct keys. A direct TPM implementation needs a defined per-encryption-root mapping and should attempt each provider once before entering the passphrase retry loop.

7. **The guide needs a threat model and a real delivery path.**

   The documented [`systemd-creds` command](https://github.com/archzfs/archzfs/blob/5c20f53a0b92a44d16a7cd2f897baf04a87f1c95/src/zfs-utils/TPM2-unlock-guide.md#L9-L25) specifies no PCR policy. That is useful for unattended reboot and protects a detached drive, but does not attest the boot chain or protect the complete machine from another boot environment. The guide should explain that boundary, PCR-policy choices, recovery, and passphrase rotation without presenting one PCR selection as universally safe.

   The examples also place the passphrase in shell history, write it temporarily in plaintext, and print it to the terminal for verification. Those patterns should not be recommended.

   Finally, this Markdown file is not copied into generated package sources, installed by `package()`, or exposed through `mkinitcpio -H zfs`, so users would never receive it.

**Preferred direction**

OpenZFS 2.4.4 already provides a [generic alternate key-loading contract](https://github.com/openzfs/zfs/blob/zfs-2.4.4/contrib/initramfs/README.md#unlocking-a-zfs-encrypted-root-via-alternate-means): provider files are conditionally copied into the initramfs, receive the encryption root and key location, and may load the key before normal interactive fallback.

I would prefer adapting that released design to mkinitcpio rather than embedding `systemd-creds` and TPM libraries in the standard ZFS hook. A small generic provider seam would add no dependency or image payload for non-users and could support TPM2, FIDO2, KMS, HSMs, or local scripts through separate opt-in providers.

The suggested LUKS sidecar is also a legitimate alternative, particularly where LUKS2 key slots, FIDO2, or its recovery tooling are desirable. It is not universally superior: it introduces another storage object plus boot ordering and mounting integration, and Arch's TPM-capable `sd-encrypt` path assumes a systemd initramfs while this hook uses BusyBox. That alternative supports keeping TPM-specific machinery outside the mandatory package, but does not invalidate the use case or all downstream integration.

A carefully gated in-tree TPM implementation could still be acceptable if the maintainers deliberately choose that maintenance burden and all correctness, packaging, and documentation issues above are resolved.

**Coordination with #655**

One maintainer-side sequencing note: we are also working on #655, which changes the `zfs-utils` package structure and metadata. Our preferred order is to land and release that work first, then rebase and complete this PR against the resulting structure; this also avoids making you redo dependency, `pkgrel`, and package-generation changes.

We expect to work on #655 soon, but cannot promise a fixed schedule. There is no need to postpone discussion or refinement of the provider design, but please coordinate with us before investing heavily in a final package-level rework. If #655 is delayed or our plans change, we can revisit the order.

**Validation before approval**

Before approval, I would expect package regeneration and source verification, a successful non-TPM `mkinitcpio -P`, inspection of the resulting image, BusyBox ash validation, and disposable boot tests covering provider success and manual-passphrase fallback. URI `keylocation` and multiple encryption roots also need tests if the direct implementation remains.

For transparency, this review draws on two model-assisted reviews conducted independently with different models, GPT-5.6 Sol and Claude Opus 5; I reconciled their findings against the PR and upstream sources.

kerberiel (changes requested):
Thanks for the thoughtful response, and for the substantial rewrite since the first review. I should have answered the design question more explicitly at the time: yes, adapting OpenZFS's alternate key-loading contract to mkinitcpio, with TPM2 implemented as an opt-in provider, is the direction I had in mind.

I agree with keeping the scope narrow. This PR does not need to become a general key-management framework, add per-dataset credential mappings, or cover FIDO2, KMS, HSMs, and similar mechanisms. The LUKS sidecar is a legitimate alternative, but it is a different design and can remain out of scope. Limiting unattended unlocking to the encryption root needed by the boot dataset is also reasonable. Using `systemd-creds` is preferable to implementing TPM operations inside ArchZFS.

The rewrite is a considerable improvement. It resolves the original checksum, package-release, unconditional payload, mandatory dependency, `systemd-analyze`, URI `keylocation`, and documentation-delivery problems. More importantly, ordinary users who do not configure a provider now receive no TPM payload or behavior change.

I am keeping `REQUEST_CHANGES`, however, because a few boot-path and packaging issues remain.

### Required Changes

1. **Run providers in a subshell.**

   [`zfs_try_load_key_provider()` currently sources each provider directly](https://github.com/archzfs/archzfs/blob/c634ab922d979c447bb9560151c930d65c59ff48/src/zfs-utils/zfs-utils.initcpio.hook#L11-L31). Mkinitcpio runs this hook in the top-level BusyBox `/init` shell, which is PID 1. A provider containing `exit` or a syntax error therefore terminates init before either the passphrase prompt or SSH fallback is reached. Provider variables and function definitions can also leak into the surrounding boot hook.

   I reproduced the termination behavior with mkinitcpio's BusyBox ash. OpenZFS deliberately uses [`(. "${key}")`](https://github.com/openzfs/zfs/blob/zfs-2.4.4/contrib/initramfs/scripts/zfs#L456-L464), which contains both failures and state changes. Please use the same subshell form and make `DATASET`, `ENCRYPTIONROOT`, `KEYLOCATION`, `ZFS`, `ZPOOL`, and the loop variable local to the provider function. I confirmed that BusyBox ash's dynamic scope still makes those variables visible inside the sourced provider.

2. **Make TPM library and module inclusion robust.**

   The install hook [hard-codes six versioned TSS sonames and adds them with `add_file`](https://github.com/archzfs/archzfs/blob/c634ab922d979c447bb9560151c930d65c59ff48/src/zfs-utils/zfs-utils.initcpio.install#L55-L63). `add_file` does not resolve ELF dependencies, while the versioned names create a standing maintenance problem whenever `tpm2-tss` changes ABI.

   Mkinitcpio's [`systemd` install hook](https://github.com/archlinux/mkinitcpio/blob/v41.1/install/systemd#L146-L166) already demonstrates the appropriate pattern: discover the relevant `libtss2-*.so*` files and use `add_binary` for real ELF objects and `add_symlink` for links. The currently required systemd TPM set is `esys`, `rc`, `mu`, and `tcti-device`; `sys` will arrive through the `esys` ELF closure, while `tctildr` is not needed for `--tpm2-device=auto`.

   Please also conditionally add `add_checked_modules '/drivers/char/tpm/'`. The common TIS and CRB drivers are built into the standard Arch kernel I inspected, but ArchZFS also supports LTS, hardened, zen, DKMS/custom kernels, and virtual or less common TPM transports. We should not make the feature depend on one kernel configuration.

3. **Use the hook's resolved boot dataset instead of parsing the command line again.**

   The TPM provider [re-parses `/proc/cmdline` using different precedence](https://github.com/archzfs/archzfs/blob/c634ab922d979c447bb9560151c930d65c59ff48/src/zfs-utils/zfs-utils.initcpio.tpm2-provider#L8-L38), while the hook [has already resolved `root=`, `zfs=`, and `bootfs`](https://github.com/archzfs/archzfs/blob/c634ab922d979c447bb9560151c930d65c59ff48/src/zfs-utils/zfs-utils.initcpio.hook#L197-L230). Conflicting or duplicate arguments can consequently make unattended unlocking decline even though the hook supports the selected dataset. This fails safely to the passphrase prompt, but it is still unnecessary feature failure in the boot path.

   Please expose the already-resolved boot dataset to providers and reduce the TPM gate to a direct comparison. This removes roughly thirty lines of subtly different parser logic and its unquoted command-substitution expansion.

4. **Tighten provider payload handling.**

   Mkinitcpio already provides `FILES=()` for additional files, so using `FILES+=(/etc/zfs/zfs-key.creds)` would be the smallest mechanism. The proposed ZFS-specific manifest has the advantage of being scoped to the ZFS hook, so keeping it is also defensible.

   If it remains, a listed but missing file should produce a build error rather than being silently omitted, and the reading loop should process a final line without a trailing newline. Silently generating an image without the credential is particularly unfriendly on a machine intended to reboot unattended.

5. **Complete and correct the setup documentation.**

   The [documented provider destination](https://github.com/archzfs/archzfs/blob/c634ab922d979c447bb9560151c930d65c59ff48/src/zfs-utils/zfs-utils.initcpio.install#L148-L179) does not exist on a clean installation. Please show a self-contained command such as `install -Dm644` rather than instructing users to copy into a missing directory.

   Please also state the `tpm2-tss` requirement, while retaining it as a direct optdepend. It currently arrives transitively through Arch's `pacman -> gnupg` dependency chain, but that unrelated dependency is not a durable interface.

   The setup example should drop `--no-tty`; it forces `systemd-ask-password` through an agent even when the administrator has an interactive terminal. The `-n` option is correct and should remain because the credential must preserve the exact key bytes.

   Please document that the credential must be recreated after `zfs change-key`, a TPM clear, or another change that rotates the TPM storage root key, and that the original passphrase must remain available out of band. It would also help to clarify that unlocking the boot dataset's inherited encryption root may unlock other datasets sharing that key.

   Finally, either document `zfs_tpm2_credential` and `zfs_tpm2_credential_name` or remove those configuration knobs. Please also state that `systemd-creds` works under the existing BusyBox initramfs and does not require adding mkinitcpio's `systemd` hook. The latter replaces BusyBox `/init`, so the existing ArchZFS runtime hook does not execute there; native systemd-initramfs support is a pre-existing limitation and is not something this PR needs to solve.

6. **Sequence this after #660.**

   The package work discussed in connection with #655 is now implemented by [#660](https://github.com/archzfs/archzfs/pull/660). Both PRs currently claim `zfs-utils 2.4.4-2`, and #660 converts the utility recipe to a split `zfs-utils`/`zfs-tests` package base.

   We tentatively expect to merge #660 over the weekend, assuming no new concern intervenes, although please treat that as an estimate rather than a firm promise. Please wait for us to confirm that it has landed, been published, and been verified before publicly rebasing this branch.

   Afterward, use the next package release, expected to be `2.4.4-3` if the upstream version has not changed. Put only the `tpm2-tss` optdepend on the `zfs-utils` child; #660 makes systemd a direct dependency already. The provider source and installed example should likewise belong only to `zfs-utils`, not `zfs-tests`.

   Please regenerate and inspect both the stable and dormant Git recipes after that rebase.

### Validation Before Final Approval

Once we agree that the code and packaging have reached a final candidate, we will also need to discuss the remaining due-diligence validation. On your side, this will likely include documenting tests of that exact revision on your real TPM configuration, primarily the setup and result of successful unattended unlocking and, if it can be tested safely, TPM or credential failure falling back to the normal passphrase prompt.

In parallel, the maintainers will inspect the resulting package and initramfs and perform our own disposable regression testing, concentrating on unchanged behavior without a configured provider and reliable manual fallback when a provider fails. As part of the QEMU coverage, we may also investigate testing with an emulated TPM.

### Final Notes

The PR description will also need to be updated once that candidate exists. It currently describes the old inline TPM pipeline, mandatory dependency, `systemd-analyze`, and standalone guide, while the attached evidence covers that superseded design. The current title is acceptable, although something like "Add a generic initramfs key-provider seam with an opt-in TPM2 example" would describe the result more precisely.

Reviewed through `c634ab922d979c447bb9560151c930d65c59ff48`.

For transparency, this review was compiled from work by GPT-5.6 Sol and Claude Opus 5. I reconciled their findings against the branch, mkinitcpio, OpenZFS, systemd documentation, current Arch package metadata, and the completed workflow run.

kerberiel (changes requested):
Thank you. The current revision resolves every substantive source change from my previous review: provider execution is isolated, the resolved boot dataset is passed directly, TPM libraries and modules follow the mkinitcpio pattern, payloads use `FILES=()`, the split-package metadata is correct, and the documentation now covers setup and recovery appropriately.

Workflow run [34152685445](https://github.com/archzfs/archzfs/actions/runs/34152685445), attempt 2, also completed the exact-head builder run successfully. It built `zfs-utils`/`zfs-tests` 2.4.4-3, `zfs-dkms`, and all four kernel families without failover package reuse. The final 403 occurred only while trying to update the shared `testing` release from a fork and is not a package-build failure.

One source correction remains. [`help()` uses an unquoted heredoc](https://github.com/radupotop/archzfs/blob/14545ed98aa9df3fa7cb1febd0dfda7a486523de/src/zfs-utils/zfs-utils.initcpio.install#L77), so `mkinitcpio -H zfs` expands the new provider variables and backslash-newlines. The rendered provider example currently contains empty values such as:

```sh
[ "" = "file" ] || return 1
keyfile=""
[ -r "" ] || return 1
"" load-key -L prompt "" < ""
```

Please change the delimiter to `cat <<'HELPEOF'` and refresh `zfs_initcpio_install_hash`. Nothing in the help body requires interpolation, so that is the complete fix. Apologies for missing this in the previous pass.

Please also refresh the PR description before the next review. It still describes the superseded inline TPM implementation, unconditional credential handling, `systemd-analyze`, mandatory `tpm2-tss`, and the removed standalone guide. Your September update comment accurately describes the current design and would make a good replacement.

I also investigated whether the TSS discovery must fail when no libraries match. I am not requesting another change there. The implementation follows mkinitcpio's established handling of these dynamically loaded libraries, `tpm2-tss` is a documented direct optdepend, and provider failure returns to the normal passphrase path.

After these two corrections, the remaining work is maintainer-side artifact inspection and bounded disposable boot validation, not another design round.

Reviewed through `14545ed98aa9df3fa7cb1febd0dfda7a486523de`.

For transparency, this review was compiled from work by GPT-5.6 Sol and Claude Opus 5. I reconciled their findings against the branch, mkinitcpio, OpenZFS, systemd source and documentation, current Arch package metadata, and the completed workflow run.

kerberiel (approved):
Thanks, this is ready from my side. I reviewed and verified the final branch through `d85933f5c001901511d730420fa78858590b1fb8`.

Maintainer-side verification included:

- Confirming that successful [Test run 34538415512](https://github.com/archzfs/archzfs/actions/runs/34538415512) built the reviewed tree from source without `failover` package reuse.
- Preserving and verifying its unsigned artifact (`sha256:5276fd3d95954cc2dff5863a5442516879222a868d0b5a0d5cf320d66cf7d583`).
- Inspecting all package archives, repository databases, `.PKGINFO`, `.BUILDINFO`, `.MTREE`, package boundaries, generated PKGBUILDs, and packaged source hashes.
- Checking the packaged scripts with Bash, BusyBox ash, and dash, including provider subshell containment.
- Building and inspecting BusyBox initramfs images with and without the provider, including the TPM payload and recursive ELF dependency closure.
- Running disposable QEMU/KVM tests with the exact standard-kernel packages and swtpm: manual passphrase unlock without a provider; unattended unlock with the enrolled TPM and no prompt; and visible TPM decryption failure with a mismatched TPM followed by successful manual-passphrase fallback.

All three boots reached the expected ZFS root with the key available and pool online. I used an explicit hostid fixture so the separate pre-existing hostid question did not influence these results.

This verification covers the exact unsigned PR artifact, one standard Arch kernel, BusyBox mkinitcpio, and an emulated TPM. It does not cover production signing/publication, physical TPM interoperability, other kernel families, PCR-bound policies, or the unsupported mkinitcpio `systemd`-hook path.

I found no remaining issue and approve this change.

For transparency, this review was compiled from work by GPT-5.6 Sol and Claude Opus 5. I reconciled their findings against the final source, exact CI artifact, package and initramfs contents, and the disposable runtime results.

PR #664, opened 2026-09-18
docs: verify exact staging synchronization

kerberizer:
## Summary

- document that remote `gh repo sync --force` can still create an upstream
  merge instead of exact branch parity
- make equality of the production and staging commit SHAs the synchronization
  gate
- provide a guarded forced-ref fallback while Actions remains disabled

## Motivation

While preserving the existing release-automation experiment in
`archzfs-testing`, GitHub CLI 2.101.0 returned success from the documented sync
command but created a merge commit. Its remote synchronization path attempts
GitHub's upstream-merge endpoint before the forced-reference fallback, so a
clean merge can bypass the expected hard reset.

The updated procedure requires independent SHA verification, repeats the
repository, archive, run, and source checks before any corrective ref move, and
keeps Actions disabled until exact parity is established.

## Validation

- `git diff --check`
- `git show --check`
- checked the documented API operation against the staging operation that
  exposed the discrepancy

This is documentation-only and does not change release, signing, package, or
workflow behavior. No additional documentation surface requires coordination.


kerberiel (approved):
Reviewed the staging synchronization safeguards and verified the GitHub CLI 2.101.0 control flow and forced-reference fallback against upstream source and API documentation. The SHA-equality gate and Actions-disabled recovery ordering are sound. `git diff --check` and `git show --check` pass; this is documentation-only, so no package build applies.

PR #665, opened 2026-09-18
docs: record PR and staging verification lessons

kerberizer:
## Summary

- require explicit verification of a long-lived pull request's stored base and displayed comparison after branch updates
- document a permissionless staging-only completion probe for `workflow_run` validation while the real mutating workflow stays disabled
- strengthen staging release snapshots, exact restoration checks, and test-workflow cleanup

## Motivation

The remediation and staging validation for #648 exposed reusable operational details that were not yet explicit in the repository documentation. This records those safeguards without carrying over case-specific commits, run IDs, or release values.

## Operational impact

Documentation only. This does not change workflows, package generation, signing, releases, or staging state.

## Validation

- `git diff --check`
- reviewed the `Test` and `Release` workflow path filters; neither changed documentation path is an input or trigger
- cross-checked the comparison-base and workflow registration guidance against
  [GitHub's pull-request comparison documentation](https://docs.github.com/en/pull-requests/reference/pull-requests#differences-between-commits-on-compare-and-pull-request-pages),
  [`workflow_run` documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run),
  and [manual workflow documentation](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow)

No package build was run because the change is limited to `AGENTS.md` and `docs/staging.md`.

## Documentation coordination

No README, release-body template, organization profile, or Wiki update is needed; this change documents repository-maintainer operating procedure only.


docs/staging.md
```diff
@@ -284,11 +306,25 @@ boundary is not a reason to grant untrusted code broader credentials.
...
-the production PR is resolved. Before unrelated work begins, repeat the
-inventory and force-sync procedure rather than accumulating experiments on
-testing `master`.
+Before release mutation, preserve the exact API representation needed to prove
+and restore the affected staging state. For each release, record its ID, tag,
+name, draft and prerelease flags, exact body string, and creation, publication,
+and update timestamps. Record the tag object and target, plus every asset's ID,
+name, size, timestamps, and available digest. Editing a fixed-name release can
+change its update timestamp without changing its publication timestamp, so
+compare both.
```
kerberiel: Pre-mutation instructions placed in the Cleanup section. This paragraph opens "Before release mutation, preserve the exact API representation", but it sits in the document's final section, so a reader following the procedure in order reaches it only after the mutation it is meant to guard.

`### 1. Inventory and Preserve State` already covers preserving branch history and says nothing about releases or tags; that is where the snapshot belongs. Suggest moving this paragraph there (or into `## Experiment Structure`) and keeping the restore/compare paragraph below in Cleanup, or at minimum adding a forward reference from section 1.

docs/staging.md
```diff
@@ -284,11 +306,25 @@ boundary is not a reason to grant untrusted code broader credentials.
...
+
+Restore release text from the captured API string rather than reconstructed
+Markdown, which may normalize or add a trailing newline. After restoration,
+compare the API representation, tag target, and asset inventory with the
+snapshot and retain any intentional discrepancy in the evidence.
+
+After verification, disable mutating and testing-only staging workflows,
+including temporary completion probes, and verify their disabled state. Ensure
+no unexpected scheduled run remains active before resetting the default branch.
```
kerberiel: Ambiguous reset. In a Cleanup section, "before resetting the default branch" can be read as the reset that ends the experiment, but from the closing paragraph it clearly means the next cycle's force-sync. Suggest "before the next force-sync", or dropping the clause since step 2 already covers it.

docs/staging.md
```diff
@@ -262,6 +262,28 @@ Keep links to the workflow runs and releases in the production PR description.
...
+and dispatch it. Preserve the production candidate separately from the overlay,
+wait for registration, resolve the probe's workflow ID, and dispatch by ID to
+avoid ambiguity when workflows share a display name. Do not treat successful
+dispatch as proof that the intended run started: identify the resulting run by
+workflow ID or path, source SHA, event, and creation time, then verify the
+downstream run and its inputs independently.
+
+Advancing staging `master` through the head of an open staging pull request can
+cause GitHub to record that pull request as merged. Treat that as an expected
+staging-side effect when applicable and record it in the experiment evidence.
```
kerberiel: Off-topic for this subsection. A pull request being auto-recorded as merged when `master` advances through its head is a general consequence of advancing the default branch, not of completion probes. It lands here because registering a probe requires pushing to the default branch, but the text never states that link.

Either make it explicit ("Because registering the probe requires advancing `master`, ...") or move it to `## Experiment Structure`.

AGENTS.md
```diff
@@ -197,6 +197,15 @@
...
   exploratory narration.
+- After updating a long-lived pull request, compare its recorded base SHA,
+  commit count, and changed-file set with a fresh three-dot comparison against
+  the current base branch. A clean local diff does not prove that GitHub has
+  refreshed the pull request's historical comparison base.
+- Changing a pull request's base, including a temporary retarget used to refresh
+  that comparison base, is a state-changing operation that requires explicit
+  authorization. It can affect comments and review state. Recheck the final
+  base and head SHAs, commits, files, review requests, and checks; delete a
+  temporary base branch only after confirming that nothing still uses it.
```
kerberiel, line 211: These two bullets split a coherent group. They are inserted between "Keep descriptions proportional..." and "Call out generated files...", which are both about writing commit and pull request descriptions and read as consecutive. Moving the pair after the "Call out..." bullet preserves that flow.

AGENTS.md
```diff
@@ -197,6 +197,15 @@
 - Keep descriptions proportional to the change. Do not restate the diff,
   document details already clear from the code, or overwhelm reviewers with
   exploratory narration.
+- After updating a long-lived pull request, compare its recorded base SHA,
+  commit count, and changed-file set with a fresh three-dot comparison against
+  the current base branch. A clean local diff does not prove that GitHub has
+  refreshed the pull request's historical comparison base.
```
kerberiel: Optional: this describes how to detect a stale comparison base but not what to do about it. The remedy is only implied by the following bullet's "temporary retarget". One clause would close the loop.

kerberiel (commented):
No blocking issues. The content is accurate and mechanically clean; every finding below is placement or wording.

## Verification

- `git diff --check` clean, no trailing whitespace, all added lines within the repository's 80-column wrap.
- Confirmed the path-filter claim independently: `test.yml` and `release.yml` both filter on `**.sh`, `.github/workflows/**`, `build-container/**`, and `packages/**`. Neither changed file matches, and GitHub reports no checks on the branch, so non-blocking treatment is correct here.
- Checked the factual claims against the live API rather than assuming. The release object does expose `created_at`, `published_at`, and `updated_at`, and assets expose `digest`, so the "compare both timestamps" and "available digest" guidance holds. `workflow_dispatch` requiring the file on the default branch, `workflow_run` matching upstream workflows by top-level `name`, and `permissions: {}` are stated accurately.
- No duplication or conflict with existing text: `master` has no prior mention of three-dot comparison, retargeting, `workflow_run`, or probes. The new cleanup text reinforces rather than contradicts the existing create-then-promote note in `AGENTS.md` about release dates.

## Note

`AGENTS.md` is canonical here and is copied unchanged into `archzfs-testing`. Nothing is required in this PR, but that copy will drift until the next sync.

AGENTS.md
kerberiel, line 211: Arguably belongs in Operational Safety. "Changing a pull request's base ... requires explicit authorization" is substantively an operational-safety rule, and that section is where the repository's other explicit-authorization rules live. Keeping pull request mechanics together here is defensible; flagging it as a discoverability judgment call, not a defect.

docs/staging.md
```diff
@@ -284,11 +314,17 @@ boundary is not a reason to grant untrusted code broader credentials.
...
 
-After verification, disable mutating staging workflows and ensure no unexpected
-scheduled run remains active. Retain the experiment branch and evidence until
-the production PR is resolved. Before unrelated work begins, repeat the
-inventory and force-sync procedure rather than accumulating experiments on
-testing `master`.
+Restore changed release text from the captured API string rather than
+reconstructed Markdown, which may normalize or add a trailing newline. After
+restoration, compare the API representation, tag target, and asset inventory
+with the snapshot and retain any intentional discrepancy in the evidence.
```
kerberiel: Optional, and a consequence of the move I asked for: with the snapshot paragraph now in step 1, this section opens on "the captured API string" and "compare ... with the snapshot" whose antecedent lives about 220 lines earlier. A short back-reference such as "the snapshot recorded in step 1" would close that.

Readability only. The instruction is still unambiguous read in isolation, so this does not need to hold the change.

kerberiel (approved):
Re-reviewed at `a1b9e78`. All actionable findings from the previous review are resolved, and nothing new is wrong.

- The release-snapshot paragraph moved to `### 1. Inventory and Preserve State`, and the reworded "Before any authorized release mutation" opener both scopes it correctly and matches the surrounding authorization language.
- The merged-pull-request note moved to `## Experiment Structure`, directly after the paragraph on advancing testing `master`. That makes the causal link self-evident without the explanatory clause I had suggested.
- The ambiguous "before resetting the default branch" clause is gone.
- `Call out generated files...` moved above the two new bullets, which keeps the description bullets together and makes the `AGENTS.md` change a clean insertion.
- "If they differ, stop and obtain explicit authorization before refreshing the comparison" closes the detect-without-remedy gap and reuses the idiom already in `docs/staging.md`.
- Leaving the base-change bullet in this section rather than Operational Safety is the better call now that the pull request mechanics sit together at the end.

## Verification

- `git diff --check` clean against `master`, no trailing whitespace, all added lines within the 80-column wrap.
- Confirmed all three moves were cuts rather than copies: `Advancing staging`, `preserve the exact API representation`, and `Call out generated files` each appear exactly once across both files. Heading structure is intact with no orphaned sections.
- Paths still do not match the `test.yml` or `release.yml` filters, and no checks are reported on the branch.
- The earlier factual checks against the live API still hold; none of that text changed.

One optional follow-on is noted inline. It is readability only and does not need to hold this change.

